PR #19, opened 2016-07-15
Implement HasMany eager loading unit tests

applenele:


Kagamine:
@applenele This unit test case name should be `HasMany eager loading test`.

PR #56, opened 2016-09-19
improve LINQ Include query performance

caleblloyd:
Relational loading from LINQ Include queries performs very poorly due to the SQL generated here: 
[IncludeExpressionVisitor.cs#L324-L366](https://github.com/aspnet/EntityFramework/blob/1.0.1/src/Microsoft.EntityFrameworkCore.Relational/Query/ExpressionVisitors/Internal/IncludeExpressionVisitor.cs#L324-L366) (see EF Issue https://github.com/aspnet/EntityFramework/issues/6569 for details)

The Exists Query Generated SQL loads all of the results from a related table into the DbDataReader.  Since our DbDataReader completes reading all data from the server when disposing, this pulls down many rows that aren't needed and takes a lot of time for data transfer.

The `innerJoinSelectExpression` is a fallback, and only loads the rows that are needed.

This PR reimplements `IncludeExpressionVisitor.cs` without the Exists Query Generation and instead relies on `innerJoinSelectExpression`, which ends up being much faster.


Kagamine:
:shipit:

Kagamine:
Published prerelease-10005

PR #71, opened 2016-10-05
Improve Sync IO Performance

### LINKED ISSUE #59: Improve Synchronous Performance
With the merging of #58, Async Performance is solid.  Threads are not held and Performance tests can run 500+ RPS while keeping response time under 1s.

There is problems with threads getting held during Synchronous performance tests, however.  Synchronous tests have trouble achieving 100 RPS without lockups.

While it is best practice to use Async methods, we can't guarantee that everyone is doing that.  Need to investigate synchronous performance issues and improve so that there are no lockups.

caleblloyd:
Fixes #59

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/Storage/IOBehavior.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Storage/IOBehavior.cs
@@ -1,0 +1,19 @@
+﻿// ReSharper disable once CheckNamespace
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+	/// <summary>
+	/// Specifies whether to perform synchronous or asynchronous I/O.
+	/// </summary>
+	internal enum IOBehavior
+	{
+		/// <summary>
+		/// Use synchronous I/O.
+		/// </summary>
+		Synchronous,
+
+		/// <summary>
+		/// Use asynchronous I/O.
+		/// </summary>
+		Asynchronous,
+	}
+}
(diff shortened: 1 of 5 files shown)


Kagamine:
Why not set async behavior as default in XXXAsync()?

caleblloyd:
Mixing Sync and Async calls in .NET [leads to problems](http://stackoverflow.com/a/24298425/1419658)

If the connection is accessed in a synchronous manner, such as in the perf test [SyncController.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/Controllers/SyncController.cs), then we need to ensure that Synchronous IO (Socket.Connect, Socket.Send, Socket.Receive) is used.

If the connection is accessed in an asynchronous manner, such as in the perf test [AsyncController](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/Controllers/AsyncController.cs), then we need to use Asynchronous IO (Socket.ConnectAsync, Socket.SendAsync, Socket.ReceiveAsync)

If we force synchronous calls down an async callpath (like we do before this PR), it can lead to Thread Pool Starvation.  We fixed all of the codepaths in MySqlConnector to either be Synchronous all the way through or Async all the way through in https://github.com/bgrainger/MySqlConnector/issues/62.  I used the same approach by defining `IOBehavior` in this PR.

Async is the best way to go, but the caller must implement async first.  We should make a "Best Practices" section of the readme or wiki and point people to the [AsyncController](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/Controllers/AsyncController.cs) as an example of the proper way to use the library.

Kagamine:
All right, :shipit:

Issue #79, opened 2016-10-10
How should we store DateTimeOffset


caleblloyd:
Reference Material:

- [DateTime vs DateTimeOffset](http://stackoverflow.com/a/14268167/1419658)
- [Storing DateTime (UTC) vs. storing DateTimeOffset](http://stackoverflow.com/a/9314893/1419658)

**Option 1: Convert the DateTimeOffset to UTC DateTime and store that**

Pros
- Sorting and comparison will work in the DBMS
- Will only take 8 bytes of storage space in DBMS

Cons
- The original TimeOffset context will be stripped
- The original timezone context will be stripped

**Option 2: Store a String Representation of DateTimeOffset**

Pros
- The original TimeOffset context will be preserved
- The original timezone context will be preserved

Cons
- Sorting and comparison will not work in DBMS
- Will take at least 33 bytes of storage space in DBMS ("2016-10-10T13:49:58.2065722+00:00")

garmiter:
Whether we can in a way compatible with two kinds of situations? Are stored in different areas, for example, decided to use which one according to the environment, or to identify different storage area according to the function.

caleblloyd:
@garmiter I don't think storing two fields is an option.  EF does a 1-to-1 mapping of properties to database fields.

The caller _can_ implement this behavior themselves by defining a DateTime field and a metadata field with a time offset / time zone.  But that is outside the bounds of this library.

garmiter:
Then use the first one(option 1), because can support the DBMS is more important, keep TimeOffset context, I think we should be able to use similar way to solve this problem.

caleblloyd:
That's also my vote.  I think we should use option 1 but flatten to a DateTime(6) in UTC instead of a long representing the timestamp in milliseconds like currently.  @Kagamine what do you think?

Kagamine:
I always avoid use DateTimeOffset in my project but Identity v3 is using type DateTimeOffset. I dont know what issue will occur when changing the store behavior. I suppose DateTime utc is ok, but need test working with Identity v3.

caleblloyd:
Added Identity V3 to the test bench in #82, but haven't written any specific test cases for them yet.

[...]

PR #83, opened 2016-10-11
Normalize DateTime Types

caleblloyd:
- CLR type `DateTime` is now stored as MySQL type `DateTime(6)`.  No time zone conversion takes place upon storing/retrieval
- CLR type `DateTimeOffset` is now stored as MySQL type `DateTime(6)`.  Converted to UTC upon storing and when retreiving
- Added test for CLR type `TimeSpan`, which maps to `Time(6)`

Now all of our Date/Time types have microsecond granularity because they are defined by default with the `(6)` argument

For reference, discussion is in #79


mylemans:
Looking forward for this one, just noticed 5 min ago that the backend field for DateTimeOffset was a varchar :P

Kagamine:
Thanks @caleblloyd

PR #101, opened 2016-10-31
Getting datetime(6) to work

alexknips:
Hi.

I have had problems using `datetime(6)` with pomelo. Our standard datetime migrations always create columns with `datetime(6)` and we like that as it is very accurate. I think this is now the new standard behaviour of Pomelo Mysql, or am I wrong?
However there are two issues

1. The default value is choosen wrong for this type of column `datetime(6)` and needs to be set similar to `property.Relational().DefaultValueSql = "CURRENT_TIMESTAMP(6)"`

2. `ValueGeneratedOnAddOrUpdate` is ignored for the column type `datetime(6)`

To help with fixing the issue I created some tests to show how I imagine it to work and a solution. 

The diff in the tests is misleading.

Thanks for your great project. Any help is appreciated.

PR #102, opened 2016-10-31
Bugfix for Computed datetime

caleblloyd:
Fixes #101

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
@@ -16,7 +17,9 @@ namespace Microsoft.EntityFrameworkCore.Migrations
 {
     public class MySqlMigrationsSqlGenerationHelper : MigrationsSqlGenerator
     {
-        public MySqlMigrationsSqlGenerationHelper(
+	    private static readonly Regex TypeRe = new Regex(@"([a-z0-9]+)\s*?(?:\(\s*(\d+)?\s*\))?", RegexOptions.IgnoreCase);
+
+	    public MySqlMigrationsSqlGenerationHelper(
             [NotNull] IRelationalCommandBuilderFactory commandBuilderFactory,
             [NotNull] ISqlGenerationHelper SqlGenerationHelper,
             [NotNull] IRelationalTypeMapper typeMapper,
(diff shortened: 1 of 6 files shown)


caleblloyd:
@Kagamine please code review

I've bumped `master` branch to `1.1.0`, and also started a `1.0.2-wip` branch

My plan is to continue new development in `1.1.0` and backport bugfixes to `1.0.2-wip` while .NET Core 1.1 is still in preview.  This way our users can use `1.0.2-prerelease` builds to get bugfixes to `1.0.1`.

src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -316,13 +317,25 @@ protected override void ColumnDefinition([CanBeNull] string schema, [NotNull] st
                 type = TypeMapper.FindMapping(property).StoreType;
             }
 
-            var generatedOnAddAnnotation = annotatable[MySqlAnnotationNames.Prefix + MySqlAnnotationNames.ValueGeneratedOnAdd];
+	        var matchType = type;
+	        var matchLen = "";
```
Kagamine: Could you use int instead of string? string may cause a sql injection.

src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -316,13 +317,25 @@ protected override void ColumnDefinition([CanBeNull] string schema, [NotNull] st
...
+	        var typeRe = new Regex(@"([a-z0-9]+)(?:\s+)?(?:\((\d+)?\))?", RegexOptions.IgnoreCase);
```
Kagamine: Pull this variable out to be a static member.

src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -338,7 +351,7 @@ protected override void ColumnDefinition([CanBeNull] string schema, [NotNull] st
                         break;
                     case "datetime":
                     case "timestamp":
-                        defaultValueSql = "CURRENT_TIMESTAMP";
+                        defaultValueSql = "CURRENT_TIMESTAMP(" + matchLen +")";
```
Kagamine: $"CURRENT_TIMESTAMPE({ mattchLen })"

src/Pomelo.EntityFrameworkCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
caleblloyd: Would have to convert it from string to int.  This is in the migration builder, not on any runtime code.  I don't think there's any issue of SQL Injection as this only takes input from EF Annotations and Fluent API during migrations, right?

caleblloyd: Also, due to the regex, match 2 will always be a number regardless

caleblloyd:
Code review changes complete

Issue #127, opened 2016-11-24
Scaffolding error w/ CASCADE referential action


mguinness:
### Steps to reproduce
A table definition that includes a referential action constraint will produce error "Sequence contains no matching element".

```
CONSTRAINT `FK_AspNetRoleClaims_AspNetRoles_RoleId` FOREIGN KEY (`RoleId`) REFERENCES `AspNetRoles` (`Id`) ON DELETE CASCADE
```

### The issue

```
System.InvalidOperationException: Sequence contains no matching element
   at System.Linq.Enumerable.Single[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.MySqlDatabaseModelFactory.GetIndexes()
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.MySqlDatabaseModelFactory.Create(DbConnection connection, TableSelectionSet tableSelectionSet)
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.MySqlDatabaseModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.EntityFrameworkCore.Scaffolding.RelationalScaffoldingModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.MySqlScaffoldingModelFactory.Create(String connectionString, TableSelectionSet tableSelectionSet)
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.ReverseEngineeringGenerator.GetMetadataModel(ReverseEngineeringConfiguration configuration)
   at Microsoft.EntityFrameworkCore.Scaffolding.Internal.ReverseEngineeringGenerator.GenerateAsync(ReverseEngineeringConfiguration configuration, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Design.Internal.DatabaseOperations.ScaffoldContextAsync(String provider, String connectionString, String outputDir, String dbContextClassName, IEnumerable`1 schemas, IEnumerable`1 tables, Boolean useDataAnnotations, Boolean overwriteFiles, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Design.OperationExecutor.<ScaffoldContextImpl>d__22.MoveNext()
[...]

PR #130, opened 2016-11-26
Database scaffolding fixes

mguinness:
For issue https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/127 and new type mappings.


src/Pomelo.EntityFrameworkCore.MySql.Design/Internal/MySqlDatabaseModelFactory.cs
```diff
@@ -27,7 +27,7 @@ public class MySqlDatabaseModelFactory : IInternalDatabaseModelFactory
         public MySqlDatabaseModelFactory(/* [NotNull] */ ILoggerFactory loggerFactory)
         {
             // Check.NotNull(loggerFactory, nameof(loggerFactory));
-
+            loggerFactory.AddConsole();
```
mguinness: I can, but when there are issues with tables (like missing PK) or fields (data type mapping) then the end user will no idea what went wrong (or what was skipped).  I used this successfully to fix problems that I encountered.  If you still feel that this is unnecessary, I will remove it.

PR #132, opened 2016-11-29
Scaffolding fixes

mguinness:
Couple of issues addressed:

- Views are now ignored for scaffolding
- Bug when --table option used when FK exists


src/Pomelo.EntityFrameworkCore.MySql.Design/Internal/MySqlDatabaseModelFactory.cs
```diff
@@ -87,7 +87,7 @@ public DatabaseModel Create(DbConnection connection, TableSelectionSet tableSele
             }
         }
 
-        const string GetTablesQuery = @"SHOW TABLES";
+        const string GetTablesQuery = @"SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'";
```
Kagamine: `TABLE_TYPE`

Issue #137, opened 2016-12-02
Scaffold mysql boolean to sbyte?


joevchs:
### Steps to reproduce
After runnning the scaffold command, the boolean type column was defined as sbyte instead of boolean on the entity model.

### The issue
The result is:
public **sbyte** tableColumn { get; set; }

### Further technical details

MySQL version: 5.7.12
Operating system: Windows 7
    "Microsoft.EntityFrameworkCore.Tools": "1.1.0-preview4-final",
    "Pomelo.EntityFrameworkCore.MySql": "1.1.0",
    "Pomelo.EntityFrameworkCore.MySql.Design": "1.1.0"

mguinness:
Please provide your table schema and confirm you're using bit datatype and not tinyint.

Kagamine:
@joevchs We will not treat DB type tinyint(1) to CLR type bool, because not all tinyint(1) are boolean. You can modify the field type after scaffolding manually.

joevchs:
Using bit(1) works as expected. Thanks!!

PR #138, opened 2016-12-03
Scaffold tinyint(1) as bool

mguinness:
To resolve issue https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/137.  Also added `tinyint unsigned` to scaffold as `byte` but it still gets defined as `sbyte` - any ideas?


Kagamine:
Could you explain me why you want to convert tinyint(1) into boolean? Is there any documents you have refered to?

mguinness:
Sorry, I should have done a better job at explaining this.  The [boolean datatype](http://dev.mysql.com/doc/refman/5.7/en/numeric-type-overview.html) in MySQL is actually a synonym for tinyint(1).

As such the ADO.NET data provider for MySQL treats it as a boolean by default and I was getting an unhandled error in the GetByte override in the SynchronizedMySqlDataReader class.  

The outermost exception was "There is already an open DataReader associated with this Connection which must be closed first" which was confusing, but the innermost exception was "Unable to Cast Object of type 'System.Boolean' to type 'System.Byte'".

The issue https://github.com/mysql-net/MySqlConnector/issues/141 will allow for the TreatTinyAsBoolean connection string option to be set.  In the meantime, I changed the underlying field to tinyint(3) to workaround the issue.

caleblloyd:
Since the ADO.NET driver's default behavior is to treat `TINYINT(1)` as `bool`, I agree that our library's default behavior should be to scaffold `TINYINT(1)` as `bool`. @Kagamine do you agree?

When using code-first, we will still use `BIT(1)` as our boolean type.

@mguinness this PR has a lot of commits (it should just have 1, reset/rebase if you can) but I think the intended change is only 2 lines: https://github.com/caleblloyd/Pomelo.EntityFrameworkCore.MySql/commit/3f9bfe8b67c167a6986f572dae237983e0120ce9 Does this look correct?  If so I can open a PR for that instead.

Really what needs to happen is Scaffolding needs tests.  Then we can fix this stuff a lot easier.

Kagamine:
boolean has been stored as tinyint(1), but not all tinyint(1) is boolean, so I dont agree.

caleblloyd:
> boolean has been stored as tinyint(1), but not all tinyint(1) is boolean, so I dont agree.

This is true, technically since `TINYINT(1)` is an 8-bit signed integer, the `(1)` in `TINYINT(1)` [only matters if zeroffill is used](http://stackoverflow.com/a/37772252/1419658)

So it seems the best path forward is to notify users to convert their `TINYINT(1)` fields that are actually booleans to `BIT(1)` fields before scaffolding then.

mguinness:
Following the decision to leave tinyint(1) scaffolding as is, I'll raise the other problems I've encountered as new issues.

Issue #142, opened 2016-12-06
Exception when TreatTinyAsBoolean option is true (default)


mguinness:
When you have a table column as `tinyint(1)` (which is scaffolded as sbyte), a LINQ query that references that will fail with InvalidCastException.

### Steps to reproduce
```SQL
CREATE TABLE TestTable (
  TestColumn tinyint(1) NOT NULL
);
```

```C#
public partial class TestTable
{
    public sbyte TestColumn { get; set; }
}
```

### The issue
`var test = context.TestTable.ToList();` will generate exception message "There is already an open DataReader associated with this Connection which must be closed first".

That exception is misleading as it's an artifact of the handled exception trying to close the transaction when the reader was still open.

The underlying exception message is "Unable to Cast Object of type 'System.Boolean' to type 'System.Byte'" which is in the GetByte override in the SynchronizedMySqlDataReader class.

### Workarounds
Changing the column from tinyint(1) to tinyint(3) will mean that the ADO.NET data provider will treat this a byte instead of a boolean.

There is an open issue https://github.com/mysql-net/MySqlConnector/issues/141 for the ADO.NET data provider.  If that is implemented, then setting `TreatTinyAsBoolean=false` in the connection string should work.

### Further technical details

MySQL version: 5.6.27-log
Operating system: Linux
Pomelo.EntityFrameworkCore.MySql version: 1.1.0

caleblloyd:
There is an open discussion in MySqlConnector at https://github.com/mysql-net/MySqlConnector/issues/141 on whether to default to `TreatTinyAsBoolean=true` or `TreatTinyAsBoolean=false`

Should we force `TreatTinyAsBoolean=false` in this library, since we store bools as `BIT(1)`?

mguinness:
It's a hard one to call.  My gut feeling is that this library should default to `TreatTinyAsBoolean=false` and also `ConvertZeroDateTime=true` (since .NET has no zero datetime).

However, if settings are overridden from the [documented](https://dev.mysql.com/doc/connector-net/en/connector-net-connection-options.html) defaults it could get confusing for end users.

Also should we see which defaults are used in Official MySql, Sapient Guardian & DevArt?  Switching providers should be transparent ideally.

caleblloyd:
> should we see which defaults are used in Official MySql, Sapient Guardian & DevArt

I don't think so.  We should decide what works best in this library and document differences from the official library.

[...]

PR #146, opened 2016-12-08
Fix for issue #142

mguinness:
Resolves following exceptions:

`There is already an open DataReader associated with this Connection which must be closed first`

`Unable to Cast Object of type 'System.Boolean' to type 'System.Byte'`

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlRelationalTransaction.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlRelationalTransaction.cs
@@ -9,6 +9,7 @@ using Microsoft.EntityFrameworkCore.Internal;
 using Microsoft.EntityFrameworkCore.Utilities;
 using Microsoft.Extensions.Logging;
 using MySql.Data.MySqlClient;
+using System.Data;
 
 namespace Microsoft.EntityFrameworkCore.Storage.Internal
 {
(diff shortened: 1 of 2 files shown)


src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlRelationalTransaction.cs
```diff
@@ -87,7 +88,7 @@ public void Dispose()
             if (!_disposed)
             {
                 _disposed = true;
-                if (_transactionOwned)
+                if (_transactionOwned && _dbTransaction.Connection.State == ConnectionState.Closed)
```
caleblloyd: This behavior is incorrect.  I know it's causing the outer exception, but this is not the appropriate fix.

caleblloyd:
I'm trying to better understand this.  Is the ADO.NET layer sending `TINYINT(1)` across as Boolean?  If this is the case, I think that the exception is appropriate and the correct fix is https://github.com/mysql-net/MySqlConnector/issues/141.

Otherwise, a `TINYINT(1)` with a value that is not 0 will always be changed to (byte) 0x01, which is NOT the correct behavior.

caleblloyd:
@mguinness let's fix this one upstream at https://github.com/mysql-net/MySqlConnector/issues/141

We should discuss forcing `TreatTinyAsBoolean=false` in this library in #142

src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlRelationalTransaction.cs
mguinness: I did look to catch the exception earlier in the call stack, but was unable to figure out a more elegant method.  Ideally, the inner exception should rollback any transactions and close the connection.

PR #156, opened 2016-12-19
[DO NOT MERGE] Smart type mapper

Kagamine:
Making type mapper auto detect the connstr and mysql version.

Related issues: #137, #145

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlTypeMapper.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlTypeMapper.cs
@@ -7,18 +7,23 @@ using System.Data;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Utilities;
+using Microsoft.Extensions.DependencyInjection;
+using MySql.Data.MySqlClient;
 
 // ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore.Storage.Internal
 {
     public class MySqlTypeMapper : RelationalTypeMapper
     {
-	    // boolean
-	    private readonly RelationalTypeMapping _bit              = new RelationalTypeMapping("bit", typeof(bool), DbType.Boolean);
+        private readonly IServiceProvider services;
+
+        // boolean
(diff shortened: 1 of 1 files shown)


src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlTypeMapper.cs
```diff
@@ -60,9 +68,27 @@ public class MySqlTypeMapper : RelationalTypeMapper
...
+            this.services = services;
+            var relationalConn = services.GetService<MySqlRelationalConnection>();
+            var TreatTinyIntAsBoolean = true;
+            var UseDateTime6 = false;
+            if (relationalConn != null)
+            {
+                var csb = new MySqlConnectionStringBuilder(relationalConn.DbConnection.ConnectionString);
+                TreatTinyIntAsBoolean = csb.TreatTinyAsBoolean;
+                relationalConn.DbConnection.Open();
+                using (var cmd = new MySqlCommand("SELECT VERSION();", relationalConn.DbConnection as MySqlConnection))
```
caleblloyd: relationalConn.DbConnection.ServerVersion should be populated after a connection has been opened [api reference](https://docs.microsoft.com/en-us/dotnet/core/api/system.data.common.dbconnection#System_Data_Common_DbConnection_ServerVersion)

We don't want to require a connection to be opened here in non-scaffolding cases though.  Any way to make the version check depend on scaffolding?

Kagamine: datetime(6) is related to migrations part.

caleblloyd: Oh, if TypeMapper is only loaded in Migrations and Scaffolding then it should be OK to open a connection here.  As long as it does not get loaded during normal operation.

Kagamine:
image

It seems that we cannot get the service while scaffolding.

caleblloyd:
> It seems that we cannot get the service while scaffolding.

I know it may be bad design, but we could create a `static` class that saves `MySqlRelationalConnection` metadata the first time a connection is opened in a process and use that.

This would not take into account multiple connection strings in the same process.  But I don't think that ever happens with scaffolding and migrations.

Kagamine:
No, it will be a bad design when using ef to connect multi databases

PR #157, opened 2016-12-20
Fix for Batch Operations

### LINKED ISSUE #151: Batching / bulk queries - MaxBatchSize()
### Steps to reproduce
	using (var dataContext = GetDataContext())
	{
		await dataContext.Database.EnsureCreatedAsync();
		
		var account = new Account { Name = "Account 1") };
		var account2 = new Account { Name = "Account 2") };
		
		dataContext.Accounts.Add(account);
		dataContext.Accounts.Add(account2);
		
		dataContext.SaveChanges();
	}

	private static ApplicationDbContext GetDataContext()
	{
		const string connectionString = "myConnectionString";

		var optionsBuilder = new DbContextOptionsBuilder<ApplicationDbContext>();
		optionsBuilder.UseMySql(connectionString, options => options.MaxBatchSize(2));
		return new ApplicationDbContext(optionsBuilder.Options);
	}
	
	public class ApplicationDbContext : BaseDbContext
    {
        public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options) : base(options) { }
	}
	
	public class BaseDbContext
    {
        // Just DbSets, entities etc.
	}
	
	public class Account
    {
        public int Id { get; set; }
	
        public string Name { get; set; }
    }

### The issue
If you specify the MaxBatchSize as an option for the DbContextOptionsBuilder the SaveChanges()-method will fail. Everything works fine if I remove the MaxBatchSize() option or if I specify it to MaxBatchSize(1). But as long as input number is > 1 it won't work.

```
Exception message: Database operation expected to affect 2 row(s) but actually affected 1 row(s). Data may have been modified or deleted since entities were loaded. See http://go.microsoft.com/fwlink/?LinkId=527962 for information on understanding and handling optimistic concurrency exceptions.

Stack trace: at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ThrowAggregateUpdateConcurrencyException(Int32 commandIndex, Int32 expectedRowsAffected, Int32 rowsAffected)
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.ConsumeResultSetWithPropagation(Int32 commandIndex, DbDataReader reader)
   at Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch.Consume(DbDataReader reader)
   at Microsoft.EntityFrameworkCore.Update.Internal.MySqlModificationCommandBatch.Execute(IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.Update.Internal.MySqlBatchExecutor.Execute(IEnumerable`1 commandBatches, IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Notified.Tests.DatabaseTests.<SetupAsync>d__6.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
   at Notified.Tests.DatabaseTests.<Test>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.TestInvoker`1.<>c__DisplayClass46_1.<<InvokeTestMethodAsync>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.ExecutionTimer.<AggregateAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Xunit.Sdk.ExceptionAggregator.<RunAsync>d__9.MoveNext()
```

caleblloyd:
- Fixes #151 
- Removes bulk insert logic.  We can't use it because we must call `LAST_INSERT_ID()` after each insert to get generated IDs.  I don't think there's any other way to do this with MySQL.
- Adds command batching test
- Adds logging to Performance / Functional tests (change to `Information` or `Debug` in `appsettings.json`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,4 +23,5 @@ script:
   - docker exec -it dotnet sh -c 'cd /dotnet && dotnet restore'
   - docker exec -it dotnet sh -c 'cd /dotnet/src/Pomelo.EntityFrameworkCore.MySql && dotnet build -c Release'
   - docker exec -it dotnet sh -c 'cd /dotnet/test/Pomelo.EntityFrameworkCore.MySql.Tests && dotnet test -c Release'
-  - docker exec -it dotnet sh -c 'cd /dotnet/test/Pomelo.EntityFrameworkCore.MySql.PerfTests && sed -i "s/127.0.0.1/mysql/g" config.json.example && dotnet test -c Release;'
+  - echo "Testing with EF_BATCH_SIZE=1" && docker exec -it dotnet sh -c 'cd /dotnet/test/Pomelo.EntityFrameworkCore.MySql.PerfTests && sed -i "s/127.0.0.1/mysql/g" config.json.example && dotnet test -c Release;'
+  - echo "Testing with EF_BATCH_SIZE=10" && docker exec -it dotnet sh -c 'cd /dotnet/test/Pomelo.EntityFrameworkCore.MySql.PerfTests && export EF_BATCH_SIZE=10 && dotnet test -c Release;'
--- a/build-run.ps1
+++ b/build-run.ps1
@@ -115,6 +115,13 @@ if ($LASTEXITCODE -ne 0){
 # run functional tests if not on MyGet
 if ($env:BuildRunner -ne "MyGet"){
     cd (Join-Path $repoFolder (Join-Path "test" "Pomelo.EntityFrameworkCore.MySql.PerfTests"))
+    echo "Testing with EF_BATCH_SIZE=1"
+    & dotnet test -c Release
+    if ($LASTEXITCODE -ne 0){
+        exit $LASTEXITCODE;
+    }
+    echo "Testing with EF_BATCH_SIZE=10"
+    $env:EF_BATCH_SIZE = 10
     & dotnet test -c Release
     if ($LASTEXITCODE -ne 0){
         exit $LASTEXITCODE;
--- a/src/Pomelo.EntityFrameworkCore.MySql/Update/Internal/MySqlModificationCommandBatch.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Update/Internal/MySqlModificationCommandBatch.cs
@@ -123,7 +123,7 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
       var grouping = UpdateSqlGenerator.AppendBulkInsertOperation(stringBuilder, _bulkInsertCommands, lastIndex);
       for (var i = lastIndex - _bulkInsertCommands.Count; i < lastIndex; i++)
       {
-        CommandResultSet[i] = grouping;
+        CommandResultSet[i] = ResultSetMapping.NoResultSet;
       }
 
       if (grouping != ResultSetMapping.NoResultSet)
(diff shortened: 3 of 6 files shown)


blixmo:
- Removes bulk insert logic. We can't use it because we must call LAST_INSERT_ID() after each insert to get generated IDs. I don't think there's any other way to do this with MySQL.

This won't solve the issue since it's the bulk inserts I need... The only reason to use the MaxBatchSize() is to activate the bulk operations, including inserts. Are you sure there is no way to retrive the generated IDs from a bulk insert with MySQL?

Maybe it should be possible to use the bulk insert anway and just don't return the generated IDs? My case is that I've a list of x tousands items and it's so ugly (and stupid) to insert all these items one by one to the database instead of all at once in one batch. There is no need to get the generated IDs back in this case.

blixmo:
- Removes bulk insert logic. We can't use it because we must call LAST_INSERT_ID() after each insert to get generated IDs. I don't think there's any other way to do this with MySQL.

This won't solve the issue since it's the bulk inserts I need... The only reason to use the MaxBatchSize() is to activate the bulk operations, including inserts. Are you sure there is no way to retrive the generated IDs from a bulk insert with MySQL? It's working for SQL Server.

blixmo:
- Removes bulk insert logic. We can't use it because we must call LAST_INSERT_ID() after each insert to get generated IDs. I don't think there's any other way to do this with MySQL.

This won't solve the issue since it's the bulk inserts I need... The only reason to use the MaxBatchSize() is to activate the bulk operations, including inserts. Are you sure there is no way to retrive the generated IDs from a bulk insert with MySQL? It's working for SQL Server.

One solution if you use INT with IDENTITY(1.1) as PK is that mysql_insert_id() will return the ID of the first inserted item. Then just increase that by 1 for every row you have inserted.

Found something here too: https://github.com/mysqljs/mysql/issues/1284

Anoter solution might be to remove the LAST_INSERT_ID() if you choose to perform a bulk insert. In my case I've a huge list of items that I want to insert as a batch. Inserting multiple items one by one is really bad performance-wise.

Kagamine:
> One solution if you use INT AUTO_INCREMENT as PK is that LAST_INSERT_ID() will return the ID of the first inserted item in the batch. Then just increase that by 1 for every row you have inserted.

I do not agree, it will not be continuation if there were concurrency an another inserting when executing the bulk insert

caleblloyd:
> This won't solve the issue since it's the bulk inserts I need...

Bulk operations are typically a bad idea to implement using an ORM.  ORMs map an object for every object inserted.  If you are iterating through a stream performing batch inserts, your heap is going to grow to be very large.  Entity Framework also has to track each object and run through logic to figure out what _can_ be batched every time it does an insert.

I would suggest that you run normal ADO.NET inserts or use [Dapper](https://github.com/StackExchange/dapper-dot-net) for large batch inserts.  If implemented properly you will see far greater performance than if using an ORM.

> The only reason to use the MaxBatchSize() is to activate the bulk operations, including inserts.

`MaxBatchSize` batches multiple statements to the database, cutting down on Round Trip Time.  So if the batch size is 10 and you have 10 different insert/update statements upon calling `db.SaveChanges`, it sends them all up at once, delimited with semicolons.  This pull request corrects the logic that makes this work, and is still a valuable addition.

caleblloyd:
@Kagamine this should be good to merge if you don't see anything that needs fixing

PR #164, opened 2016-12-21
Translations for DATE_ADD() and EXTRACT()

mguinness:
Implemented DATE_ADD function:

`var test = ctx.Post.Select(p => p.CreatedOn.AddYears(-1));`
will result in
`SELECT DATE_ADD(p.CreatedOn, INTERVAL -1 YEAR) FROM Posts AS p`

Implemented EXTRACT function:

`var test = ctx.Post.Select(p => p.CreatedOn.Year);`
will result in
`SELECT EXTRACT(YEAR FROM p.CreatedOn) FROM Posts AS p`

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/Query/ExpressionTranslators/Internal/MySqlDatePartTranslator.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/Query/ExpressionTranslators/Internal/MySqlDatePartTranslator.cs
@@ -11,18 +11,15 @@ using Microsoft.EntityFrameworkCore.Query.Expressions.Internal;
 // ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore.Query.ExpressionTranslators.Internal
 {
-    public class MySqlDateAddTranslator : IMethodCallTranslator
+    public class MySqlDatePartTranslator : IMemberTranslator
     {
-        public virtual Expression Translate([NotNull] MethodCallExpression methodCallExpression)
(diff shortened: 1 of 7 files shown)


src/Pomelo.EntityFrameworkCore.MySql/Query/ExpressionTranslators/Internal/MySqlDatePartTranslator.cs
```diff
@@ -0,0 +1,48 @@
...
+            string datePart;
+            if (memberExpression.Expression.Type == typeof(DateTime)
+                && (datePart = GetDatePart(memberExpression.Member.Name)) != null)
+            {
+                return new DatePartExpression(datePart, memberExpression.Type, memberExpression.Expression);
+            }
+            return null;
+        }
+
+        private static string GetDatePart(string memberName)
```
Kagamine: Duplicated of the above one.

mguinness: The GetDatePart is subtlety different between classes on the case statements:

`case nameof(DateTime.AddYears):` -vs- `case nameof(DateTime.Year):`

Please let me know if I misunderstood your question.  Thanks.

Kagamine: Oh yes, so I think it is ok to merge

Issue #141, opened 2016-12-06
Scaffolding unsigned tinyint


mguinness:
Scaffolding is incorrectly mapping `tinyint(x) unsigned` to sbyte instead of byte.

### Steps to reproduce
```SQL
CREATE TABLE TestTable (
  TestColumn tinyint(3) unsigned NOT NULL
);
```

Scaffolding produces:

```C#
public partial class TestTable
{
    public sbyte TestColumn { get; set; }
}
```

### The issue
After scaffolding, result should be:

```C#
public partial class TestTable
{
    public byte TestColumn { get; set; }
}
```

PR #168, opened 2017-01-02
Fixed #163 According to conn str to adjust TypeMapper

### LINKED ISSUE #163: Expose Connection String Builder to DI for use in Type Mapper
Right now we create a Connection String builder in [UseMySql()](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/3b600717db2710eec0c3da3c316d5e8c3971fe13/src/Pomelo.EntityFrameworkCore.MySql/Extensions/MySqlDbContextOptionsExtensions.cs#L25-L29).

Expose this Connection String Builder via DI so that `TypeMapper` can make use of it.  Adjust TypeMapper types based off settings for:

- TreatTinyIntAsBool in #153 
- OldGuids in #159

Kagamine:
Now, the lib could detect the conn str to treat tinyint as boolean or sbyte. But we still have a problem.

The type mapper is no longer to be singleton, in other words, each injection will generate a new object. So I wanna know where scene will inject the type mapper. Will scoped type mapper influence efficiency?


src/Pomelo.EntityFrameworkCore.MySql/Extensions/MySqlDbContextOptionsExtensions.cs
```diff
@@ -87,5 +90,25 @@ private static MySqlOptionsExtension GetOrCreateExtension(DbContextOptionsBuilde
                 ? new MySqlOptionsExtension(existing)
                 : new MySqlOptionsExtension();
         }
+
+        private static bool IsAbleToUseDateTime6(string connectionStr)
```
caleblloyd: This will cause an extra DB Connection for every DbContext that is instantiated.  We should add caching.

Kagamine: No, this is only be invoked at you call UseMySql("xxx")

Kagamine: Now we have an issue, when you `dotnet ef migrations add xxx`, the ef will lost the extension, so the UseDateTime6 and TreatTinyAsBoolean will be false. That is why the ci fails

Kagamine: I don't have a good idea to solve this problem, maybe we must use a static class to store the UseDateTime6 and TreatTinyAsBoolean

caleblloyd: I've got some time to look into this today.  Let me take a shot at adding some tests around what you have and I'll push to this PR.

Kagamine: If we add a static class, we can remove `MySqlDesignTimeTypeMapper`. The static class should contains a property `IsCached`, if it is false, then call IsAbleToUseDateTime6. The type mapper maybe should add as Transient

Kagamine: The microsoft folks gave us a suggestion: https://github.com/aspnet/EntityFramework/issues/7275

caleblloyd: I like the Microsoft suggestion, let me try to implement that.

For the caching, we can use a `ConcurrentDict` keyed off the Connection String, like MySqlConnector does with the [connection pool](https://github.com/mysql-net/MySqlConnector/blob/master/src/MySqlConnector/MySqlClient/ConnectionPool.cs#L118-L131)

Kagamine: But how to get connstr in typemapper

caleblloyd: I think the Microsoft suggestion should allow us to get it.  Trying it out now

caleblloyd:
This is close but still a few issues.  Do not merge yet, I'll post back when it's ready for review.

caleblloyd:
I added a commit to this PR that implements a scoped TypeMapper similar to the one described in https://github.com/aspnet/EntityFramework/issues/7275

It also fixes a lot of problems with signed/unsigned integers in scaffolding such as #141 

- MySql Version for `datetime` vs `datetime(6)`
- `TreatTinyAsBoolean` will scaffold `tinyint(1)` as Boolean
- `OldGuids` will use `binary(16)` for GUIDs in Code First and Scaffolding

PR #170, opened 2017-01-03
Fixed #169 set AllowUserVariables&UseAffectedRows with an opened connection

### LINKED ISSUE #169: UseMySql( DbConnection ) extension expects the connection to be closed
In MySqlDbContextOptionsExtensions.UseMySql extention method that takes a DbConnection, the code below expects (demands actually) that the connection be closed

            var csb = new MySqlConnectionStringBuilder(connection.ConnectionString)
            {
	            AllowUserVariables = true,
	            UseAffectedRows = false
            };

            connection.ConnectionString = csb.ConnectionString;

An Exception is thrown when the connection string is is set on the open connection.

 https://docs.microsoft.com/en-us/ef/core/saving/transactions 

This article explains how to share connections and transactions. This technique is not possible with your MySQL implementation.

Thanks
Jim

Kagamine:


src/Pomelo.EntityFrameworkCore.MySql/Extensions/MySqlDbContextOptionsExtensions.cs
```diff
@@ -45,13 +45,22 @@ public static class MySqlDbContextOptionsExtensions
...
-            {
-	            AllowUserVariables = true,
-	            UseAffectedRows = false
-            };
-
-            connection.ConnectionString = csb.ConnectionString;
+            var csb = new MySqlConnectionStringBuilder(connection.ConnectionString);
+	        if (csb.AllowUserVariables != true || csb.UseAffectedRows != false)
+	        {
+	            try
```
caleblloyd: I switched this to `try/catch` because even if a connection is Closed, setting the connection string will still fail if the connection has previously been opened.

Kagamine: agree

caleblloyd:
LGTM

PR #206, opened 2017-03-08
Prepare for 1.1.1

caleblloyd:
- Upgrade Entity Framework and .NET dependencies to 1.1.1
- Switch from `project.json` to `msbuild`


Kagamine:
build.ps1 depends on project.json

caleblloyd:
I updated KoreBuild to `1.0.0-msbuild-rtm`, this works with msbuild

The problem that it is hitting right now is that `test\Pomelo.EntityFrameworkCore.MySql.PerfTests` is configured to be a test project and a console project.  I don't think this will work with msbuild, I think I'll need to split that out to Performance tests and Functional tests

caleblloyd:
Got CI to build.  There's some considerations here:

- `msbuild` tooling is brand new GA and only works in VS 2017 and VS Code
- Other tools such as Project Rider do not fully support `msbuild` yet
- We could make a branch called `project_json` and people who want to contribute that don't have `msbuild` tools working yet could work on that.  It would be a cut of what master is at right now.  This would be a temporary thing for the next couple months until IDEs catch up

Overall, Microsoft has stated that `msbuild` is the way forward, so we should move towards it.

There's a fair amount of unknown changes with the dotnet ecosystem here, so I think we should publish another prerelease and have the community vet it out for a couple weeks before we tag `1.1.1`

Kagamine:
approve

caleblloyd:
> I don't think this will work with msbuild, I think I'll need to split that out to Performance tests and Functional tests

I got it to work with msbuild using the `<StartupObject>` option, so no need to split out the projects

caleblloyd:
Published in [Pomelo.EntityFrameworkCore.MySql/1.1.1-prerelease-10020](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql/1.1.1-prerelease-10020) and [Pomelo.EntityFrameworkCore.MySql.Design/1.1.1-prerelease-10020](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql.Design/1.1.1-prerelease-10020)

@Kagamine I don't think our automated tests do scaffolding yet.  Can you try scaffolding on this new prerelease and ensure that it didn't break anything?

PR #216, opened 2017-03-16
change datetime(6) to datetime

moemura:
Generate datetime(6) will fail when assigned default value
CURRENT_TIMESTAMP. Also change ALTER COLUMN to MODIFY COLUMN


moemura:
I am sorry I don't know it was duplicate.

Issue #221, opened 2017-03-20
An asynchronous socket operation is already in progress using this SocketAsyncEventArgs instance.


jp7677:
Hi there,

We have a pretty straight forward ASP .Net Core application with EF Core that does quite some we api calls with database actions concurrently. We are also do make heavy use of a PaginaList class like described here https://docs.microsoft.com/en-us/aspnet/core/data/ef-mvc/sort-filter-page#add-paging-functionality-to-the-index-method.

Usage looks like:
```
using (var dataContext = dataContextFactory.Create())
{
  var query = dataContext.Entity
  .Include(x => x.SomeSubEntity)
  .ThenInclude(x => x.SomeField)
  .ThenInclude(x => x.SomeField2)
  .Include(x => x.SomeSubEntity2)
  .ThenInclude(x => x.SomeField)
  .ThenInclude(x => x.SomeField2)
  .ThenInclude(x => x.SomeField3)
  .AsQueryable();

  query = AddListFiltering(query, filter);
  query = AddListSorting(query, filter);

  return await PaginatedList<Entity>.CreateAsync(query, filter.Page ?? DefaultPage, filter.PageSize ?? DefaultPageSize);
}

```

Unfortunately we are seeing the following exception:

### The issue
```
Exception message: An exception occurred in the database while iterating the results of a query.
Logger: Microsoft.EntityFrameworkCore.Query.Internal.MySqlQueryCompilationContextFactory
Stack trace:
System.InvalidOperationException: An asynchronous socket operation is already in progress using this SocketAsyncEventArgs instance.
   at System.Net.Sockets.SocketAsyncEventArgs.StartConfiguring()
   at System.Net.Sockets.SocketAsyncEventArgs.SetBufferInternal(Byte[] buffer, Int32 offset, Int32 count)
   at MySql.Data.Protocol.Serialization.SocketByteHandler.<DoWriteBytesAsync>d__4.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at MySql.Data.ValueTaskExtensions.ContinueWith[T,TResult](ValueTask`1 valueTask, Func`2 continuation)
[...]

PR #226, opened 2017-03-23
Buffered Result Sets

caleblloyd:
- Forces `BufferResultSets=true` introduced in MySqlConnector 0.15.1 (https://github.com/mysql-net/MySqlConnector/issues/202)
- No longer need locking while result set is being consumed


caleblloyd:
This may fix the issue in #221, we will need to ship it in a prerelease and try

PR #258, opened 2017-04-13
MySQL 顺序Guid算法更新

yuzukwok:
不同数据库顺序guid要求不同

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/ValueGeneration/Internal/MySqlSequentialGuidValueGenerator.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/ValueGeneration/Internal/MySqlSequentialGuidValueGenerator.cs
@@ -1,0 +1,196 @@
+﻿using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.ValueGeneration;
+using System;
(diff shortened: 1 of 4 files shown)


caleblloyd:
Why use this over the existing [SequentialGuidValueGenerator.cs](https://github.com/aspnet/EntityFramework/blob/dev/src/EFCore/ValueGeneration/SequentialGuidValueGenerator.cs) in upstream?

caleblloyd:
Ok, makes sense if SQL Server is putting their sequential values at the end.  MySql needs them at the beginning.

Let me see if I can fix this up a little, add tests, and open a separate PR.

yuzukwok:
thank you

caleblloyd:
We'll need to take into account GUID format in [MySqlScopedTypeMapper.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlScopedTypeMapper.cs#L78-L85)

It could be either `CHAR(36)` or `BINARY(16)`.  Will need to write a method to convert Binary to Hex and add dashes if it is `CHAR(36)`

caleblloyd:
Closing in favor of #260

PR #260, opened 2017-04-15
Sequential GUIDs

caleblloyd:
- Adapt Sequential GUID generation to work on MySQL
- Re-work of initial PR #258 
- Must take into consideration if GUIDs are stored as `binary` (`OldGuids=true`) or `char(36)` (`OldGuids`=false)

Still need to add tests before we can merge

### DIFF
--- a/src/Pomelo.EntityFrameworkCore.MySql/ValueGeneration/Internal/MySqlSequentialGuidValueGenerator.cs
+++ b/src/Pomelo.EntityFrameworkCore.MySql/ValueGeneration/Internal/MySqlSequentialGuidValueGenerator.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Security.Cryptography;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.Storage.Internal;
+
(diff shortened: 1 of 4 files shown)


jp7677:
Interesting.. What would I have to do to make advantage of this? We already use `char(36)` as column type for our GUIDs. Would this affect existing data?

caleblloyd:
This only affects GUIDs that are part of the primary key that have not been set. It will set them automatically, no data annotation needed.

This happens right now, except the generated GUID is optimized for SQL Server, and not MySQL.

jp7677:
Thanks for the info and nice improvement!

caleblloyd:
@Kagamine tests are done for this, it's ready for review

jp7677:
@Kagamine, Nice to see that this one has been merged. May I ask for a new nuget beta build, would love to play with it?

Kagamine:
myget feed 1.1.2

jp7677:
Thanks, that will do.

Kagamine:
@caleblloyd Please revert this, I got a duplicate GUID in a concurrency scene.

caleblloyd:
How long is your field? Are you truncating the random portion if the GUID?

This is a good addition, we should debug it not revert.

PR #267, opened 2017-05-03
New expression translators

mguinness:
Added math functions to support Spherical Law of Cosines calculation.  Currently this causes a table scan due to the functions being evaluated locally.


caleblloyd:
Looks like CI is failing- can you fix it or do you need me to take a look?

mguinness:
@caleblloyd When you have time, can you take a look?  The error showing is for the MySqlMathAcosTranslator test with the following error:

`System.NullReferenceException : Object reference not set to an instance of an object.`

It seems that the FirstOrDefaultAsync() is returning null.  Is there a way to rerun the tests?

mguinness:
The failing test was due to an out of range parameter for the Acos function which needs to be between -1 and 1.  Did I miss the 1.1.2 build?

Issue #288, opened 2017-06-10
2.0.0  Functional Tests update


caleblloyd:
Lots has changed in upstream 2.0.0.  The `dev` branch now tracks `2.0.0-preview3-*` from aspnet/EntityFramework's MyGet feed.

Functional tests have stopped working because migrations do not work.

Investigate cause of migration failures and get all functional tests working.

caleblloyd:
`dotnet ef migrations add initial` is failing to map `JsonObject<>` properties.  Opened upstream https://github.com/aspnet/EntityFramework/issues/8814

smitpatel:
This call is bit ambiguous

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/Pomelo.EntityFrameworkCore.MySql/Storage/Internal/MySqlTypeMapper.cs#L251

It is possible that it worked in previous versions of netstandard but not allowed in netstandard2.0 anymore.

The `PropertyMappingValidationConvention` comes through this path only. Make sure that `FindMapping` function given a generic `JsonObject` returns the type correctly. When TypeMapper doesn't return a type, the convention will fail.

caleblloyd:
@smitpatel thanks for pointing that line out,  I had erroneously removed it in the 2.0.0 migration.  Adding it back fixed the issue

caleblloyd:
Running into a problem with DI Resolving `IDBContextOptions`.  Asked for help upstream at https://github.com/aspnet/EntityFramework/issues/8818

Kagamine:
Could you add EFCore, EFCore.Relational into dev branch for debugging, I've tried to add them but cannot hit the break point.

caleblloyd:
Let me try the suggestions in https://github.com/aspnet/EntityFramework/issues/8818 first and see if I can get it working.  I need a few days before I can dig into it

Kagamine:
Hi @caleblloyd, could you give me the myget owner permission. Now we have another project `RedisConnector` with full async implements and based on libuv.

caleblloyd:
@Kagamine I'm not sure how to do that, I don't know if I have owner permission on MyGet.  Don't you already have owner permissions on there?

Kagamine:
I have transfered the ownership by mistake prrvious

caleblloyd:
Ah MyGet only allows 1 owner.  I transferred it back to you.

Kagamine:
[...]

PR #293, opened 2017-06-16
Fixes migration unit tests

Kagamine:


Kagamine:
@caleblloyd Did the UT in your machine can pass 43 cases? or you can only pass 7 cases now?

caleblloyd:
I think I need to switch the Connection Settings to do a lazy initialize

Kagamine:
In this way, the migrations will be mitigated now, but on update will be broken. I will keep investigate the next days. I was too busy in previous term to maintain this project. Thanks for your contributions these days.

caleblloyd:
@Kagamine switched `ConnectionSettings` to lazy initialize in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/commit/2ea2961e2b21604d8fe920d3f1a0bd365815a76a you can merge the dev branch to get the fix.

Now the dev branch UT is:

```
Total tests: 57. Passed: 43. Failed: 13. Skipped: 1.
```

Kagamine:
@caleblloyd Now the UT fails 3 cases only. I want to checkin this and fix them later/

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -389,10 +390,15 @@ protected override void ColumnDefinition([CanBeNull] string schema, [NotNull] st
...
-                            throw new InvalidOperationException(
-                                $"Error in {table}.{name}: DATETIME does not support values generated " +
-                                "on Add or Update in MySql <= 5.5, try explicitly setting the column type to TIMESTAMP");
+                        try {
+                            if (_options.ConnectionSettings.ServerVersion.SupportsDateTime6)
+                                throw new InvalidOperationException(
+                                    $"Error in {table}.{name}: DATETIME does not support values generated " +
+                                    "on Add or Update in MySql <= 5.5, try explicitly setting the column type to TIMESTAMP");
+                        }
+                        catch {
```
caleblloyd: We should not catch and ignore this. Instead we should mock ConnectionSettings in the UT so it doesn't fail

Kagamine: reverted

Kagamine:
Removed try catch will throw an exception, could you investigate this? @caleblloyd

caleblloyd:
Yes I'll work up a mock options

Kagamine:
@caleblloyd Does pomelo 2.0 not work?

I've got an exception:

Unhandled Exception: System.ArgumentNullException: Value cannot be null.
Parameter name: contextOptions

caleblloyd:
@Kagamine that's right it does not work yet.  I'm tracking issues with upstream as I hit them in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/288

Issue #320, opened 2017-07-12
Issue with migrations on MariaDB 10.2


nover:
### Steps to reproduce
On an existing EF model with a foreign key to one table, change that foreign key to another table - this caused EF to create a "RENAME INDEX" in my case as the column name changes.

### The issue
When renaming indexes MySQL 5.7 introduced a new nice keyword `RENAME INDEX TO` - however, Mariadb 10.2 does _not_ support this, so the generated migration fail. 

I was hoping that the MySQL connector could see / guess that it's not talking to MySQL 5.7 and create a migration that would apply cleanly - something like `DROP` and `CREATE` index though it is a slower.

Doctrine - an ORM for PHP had a similar issue:
https://github.com/doctrine/migrations/issues/196

### Further technical details

MySQL version:  MariaDB 10.2.6
Operating system: Windows 10 with database via Docker Compose
Pomelo.EntityFrameworkCore.MySql version: 1.1.2

Now I know that the driver says it is MySQL compatible, but given that MariaDB is so close to being binary compatible with MySQL I was hoping you would look into fixing this.

Kagamine:
Yes, this was by design. rename operations only works with 5.7+. We do not want to drop and re-create.

nover:
Okay, so the driver basically supports MYSQL 5.7, with no intention to support older versions of MYSQL nor mariadb?

caleblloyd:
> Okay, so the driver basically supports MYSQL 5.7, with no intention to support older versions of MYSQL nor mariadb?

No

---

The migration operation is [here in MySqlMigrationsSqlGenerator.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/e53d8fe26c0e0bfbb184b595dcb2625e6a7d17e5/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs#L154-L171)

A server version flag needs to be added [here in ServerVersion.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/e53d8fe26c0e0bfbb184b595dcb2625e6a7d17e5/src/EFCore.MySql/Storage/Internal/ServerVersion.cs) for `SupportsRenameIndex`

Then the migration option needs to check `_options.ConnectionSettings.ServerVersion.SupportsRenameIndex` and use it if available, otherwise fallback to drop and re-create.

---

[...]

PR #321, opened 2017-07-13
Add rename support for MySQL versions below 5.7 (and MariaDB)

nover:
This PR adds support for renaming indices  on MySQL versions below 5.7 and MariaDB

- `ServerVersion` updated to determine whether renaming is supported
- `MySqlMigrationsSqlGenerator` updated to support `RENAME` and `DROP+CREATE`
- Added tests for rename support on MySQL 5.6 and 5.7
- Other small changes to allow unit-testing the rename support

**one caveat** if the index being renamed is unique and it is not version 5.7 I can't se a way to determine whether the index being changed is unique or not, hence it will DROP and create a non unique index.

Relates to #320

### DIFF
--- a/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
+++ b/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
@@ -158,15 +158,36 @@ namespace Microsoft.EntityFrameworkCore.Migrations
 
             if (operation.NewName != null)
             {
-                builder.Append("ALTER TABLE ")
-                    .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
-                    .Append(" RENAME INDEX ")
-                    .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name))
-                    .Append(" TO ")
-                    .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.NewName))
-                    .AppendLine(";");
+                if (_options.ConnectionSettings.ServerVersion.SupportsRenameIndex)
+                {
+                    builder.Append("ALTER TABLE ")
+                        .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+                        .Append(" RENAME INDEX ")
+                        .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name))
+                        .Append(" TO ")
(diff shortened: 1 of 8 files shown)


caleblloyd:
> one caveat if the index being renamed is unique and it is not version 5.7 I can't se a way to determine whether the index being changed is unique or not, hence it will DROP and create a non unique index

I'll see what I can do about that

caleblloyd:
Thanks for the PR!  The MySQL 5.6 test is especially nice!

nover:
Fantastic :) I was poking around in the `IModel` that is injected which does allow to find all indexes defined on the entity, but I couldn't find an index name and gave up.

I'll create a version of this for 1.1 as well.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -158,15 +158,36 @@ protected override void Generate(RenameIndexOperation operation, IModel model, M
...
+                        .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+                        .Append(" DROP INDEX ")
+                        .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name))
+                        .AppendLine(";");
+
+                    EndStatement(builder);
+                    
+                    builder.Append("ALTER TABLE ")
+                        .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Table, operation.Schema))
+                        .Append(" CREATE INDEX ")
```
caleblloyd: I just realized this won't work; it doesn't define any column names.  I'm re-working this function so that it can detect `UNIQUE`, `BTREE`, etc. using the `SHOW CREATE TABLE` logic anyways though.

nover: Oh, yeah, you're completely right, my bad. Disregard #322 then.

mc0re:
My version string on the server is:
```
$ mysql -V
mysql  Ver 15.1 Distrib 10.4.12-MariaDB, for Linux (x86_64) using readline 5.1
```

But the index renaming is still not supported:
```
 MySql.Data.MySqlClient.MySqlException (0x80004005): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to u
se near 'INDEX `IX_CourseFiles_Hash` TO `IX_Files_Hash`' at line 1
```

Can I force the server version somehow?

PR #345, opened 2017-08-06
Support scaffold timestamp

Kagamine:
### For Issue

Scaffolding a table which contains a timestamp typed column, the field will be ignored due to timestamp is missing in our type mapper.

### Code changes

- Add timestamp into type mapper


src/EFCore.MySql/Storage/Internal/MySqlTypeMapper.cs
```diff
@@ -49,7 +49,8 @@ public class MySqlTypeMapper : RelationalTypeMapper
 	    // DateTime
         private readonly MySqlDateTimeTypeMapping _dateTime6              = new MySqlDateTimeTypeMapping("datetime(6)", DbType.DateTime);
         private readonly MySqlDateTimeOffsetTypeMapping _dateTimeOffset6  = new MySqlDateTimeOffsetTypeMapping("datetime(6)", DbType.DateTime);
-        private readonly TimeSpanTypeMapping _time6                  = new TimeSpanTypeMapping("time(6)", DbType.Time);
+        private readonly MySqlDateTimeOffsetTypeMapping _timeStamp        = new MySqlDateTimeOffsetTypeMapping("timestamp", DbType.DateTime);
```
caleblloyd: Should be `timestamp(6)`

caleblloyd: It appears to be supported: https://dev.mysql.com/doc/refman/5.7/en/fractional-seconds.html

PR #395, opened 2017-10-10
Check the connection type

scheshan:
When use a WrapperConnection, a NullReferenceException will be thrown. So I think it is better to check if the Connection/Transaction is actually the MySqlConnection/MySqlTransaction.


caleblloyd:
Thanks for the contribution!

PR #404, opened 2017-10-16
Fix optimistic concurrency

### LINKED ISSUE #403: DbUpdateConcurrencyException not being thrown in rare circumstances
### Steps to reproduce
Can only seem to reproduce this in my solution. I've tried making a small reproduction of it without success.

### The issue
When marking an entity column as a concurrency token and then simulating a concurrency conflict the call to `SaveChangesAsync()` fails to throw a `DbUpdateConcurrencyException`.

### Further technical details

MySQL version: **5.7**
Operating system: **Windows 10 Pro**
Pomelo.EntityFrameworkCore.MySql version: **2.0.0**

Not using EF migrations, but rather creating database schema manually. The error occurs when I try to update an entity that has a composite primary key composed of two columns/properties. The entity in question has a collection of child entities and each one of those child entities has it's own collection of child entities. The scenario that I see this on is when the the update includes the addition of 3 child entities in the lowest level of the hierarchy as well as making a change to a string property on the root entity.

shaynevanasperen:
Fixes #403

This fixed the issue in my solution, however I can't seem to write a unit test that fails without the fix. Some help would be appreciated.

### DIFF
--- a/src/EFCore.MySql/Update/Internal/MySqlUpdateSqlGenerator.cs
+++ b/src/EFCore.MySql/Update/Internal/MySqlUpdateSqlGenerator.cs
@@ -72,6 +72,7 @@ namespace Microsoft.EntityFrameworkCore.Update.Internal
                 if (readOperations.Length == 0)
                 {
                     AppendSelectAffectedCountCommand(commandStringBuilder, name, schema, commandPosition);
+                    resultSetCreated = true;
                 }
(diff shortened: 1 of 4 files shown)


caleblloyd:
This may already be fixed by #407 in the latest prerelease.  It will ultimately be revisited in #411 for 2.0.1.

Issue #405, opened 2017-10-17
DbUpdateException when adding entities with parent / child relationship and a many-many relationship


neoGeneva:
### Steps to reproduce
Example project included here: https://github.com/neoGeneva/PomeloDbUpdateException

Create a data model that has three entities:
1. One with a primary key (e.g. `Category`)
2. One with a parent / child relationship (e.g.`Product`)
3. One that creates a many to many relationship between the first two entities (e.g. `ProductCategory`)

### The issue
When adding multiple instances of `Product` and `ProductCategory` I'm getting a `DbUpdateException` complaining that an instance with that primary key already exists. I've executed the same code against SqlServer with Microsoft.EntityFrameworkCore.SqlServer@2.0.0 and the issues doesn't exist there.

I've done some debugging and it looks like the SQL that executes to insert the `ProductCategory` and executes `SELECT ROW_COUNT()` is assumed to be a `ProductId` by `Microsoft.EntityFrameworkCore.Update.AffectedCountModificationCommandBatch::Consume` because it doesn't move the `DbDataReader` to the next result when the mapping is `ResultSetMapping.NoResultSet`

### Exception
```
Exception message:  The instance of entity type 'Product' cannot be tracked because another instance with the key value 'ProductId:1' is already being tracked. When attaching existing entities, ensure that only one entity instance with a given key value is attached.
Stack trace:
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.IdentityMap`1.Add(TKey key, InternalEntityEntry entry)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.UpdateIdentityMap(InternalEntityEntry entry, IKey key)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.NavigationFixer.KeyPropertyChanged(InternalEntityEntry entry, IProperty property, IReadOnlyList`1 containingPrincipalKeys, IReadOnlyList`1 containingForeignKeys, Object oldValue, Object newValue)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.InternalEntityEntryNotifier.KeyPropertyChanged(InternalEntityEntry entry, IProperty property, IReadOnlyList`1 keys, IReadOnlyList`1 foreignKeys, Object oldValue, Object newValue)
[...]

Issue #397, opened 2017-10-11
The instance of entity type 'Review' cannot be tracked 


JonPSmith:
### Steps to reproduce
I am writing a book [Entity Framework in Action](http://bit.ly/2m8KRAZ) for Manning and I am trying to use MySql, but I get an error on a standard setup that has worked with SQL Server and SQLite. 

I am using Pomelo.EntityFrameworkCore.MySql version 2.0.1-preview-10068 with EF Core 2.0.0

### The issue
I want to use another database than SQL Server, and I chose MySQL because my hosting service supports that.

I started with some basic unit tests to check everything was OK. I then ran a method to seed the database,  `SeedDatabaseFourBooks`, that I have used a lot (60+ references in my unit tests) and it failed with the exception.

To try an pinpoint the problem I have copied the `SeedDatabaseFourBooks` method core into the unit test you see below. I have removed one book and it still fails - if I comment out any of the three books it works. But with the statements as shown it errors with the exception shown after the code:

```c#
[Fact]
public void TestMySqlCreateBooks()
{
    //SETUP
    var connection = AppSettings.GetConfiguration()["MySql-Test"];
    var optionsBuilder =
        new DbContextOptionsBuilder<EfCoreContext>();
    optionsBuilder.UseMySql(connection);
    optionsBuilder.EnableSensitiveDataLogging();

    using (var context = new EfCoreContext(optionsBuilder.Options))
    {
        context.Database.EnsureCreated();
        var logIt = new LogDbContext(context);
        var existingBooks = context.Books.Count();

        //ATTEMPT
        var martinFowler = new Author
        {
            Name = "Martin Fowler"
        };

        var books = new List<Book>();

        var book2 = new Book
        {
            Title = "Patterns of Enterprise Application Architecture",
            Description = "Written in direct response to the stiff challenges",
            PublishedOn = new DateTime(2002, 11, 15),
            Price = 53
        };
        book2.AuthorsLink = new List<BookAuthor> { new BookAuthor { Author = martinFowler, Book = book2 } };
        books.Add(book2);

        var book3 = new Book
        {
            Title = "Domain-Driven Design",
            Description = "Linking business needs to software design",
            PublishedOn = new DateTime(2003, 8, 30),
            Price = 56
        };
        book3.AuthorsLink = new List<BookAuthor> { new BookAuthor { Author = new Author { Name = "Eric Evans" }, Book = book3 } };
        books.Add(book3);

[...]

Issue #379, opened 2017-09-06
Generated Insert command has wrong id of parent entity


sm-g:
### Steps to reproduce
Use https://github.com/sm-g/EFCoreInMemorySample
Create db using ([schema.sql](https://github.com/sm-g/EFCoreInMemorySample/blob/master/schema.sql)), 
update [connection string](https://github.com/sm-g/EFCoreInMemorySample/blob/master/WebApp/Startup.cs#L33)
run app, call http://localhost:5501/api/sampledata/test 
[This inserts](https://github.com/sm-g/EFCoreInMemorySample/blob/master/WebApp/Controllers/SampleDataController.cs#L21) will be executed. 

### The issue
After upgrade to EF Core 2 `SELECT ROW_COUNT();` affects following `LAST_INSERT_ID()`.

mysql log file:
```
2017-09-06T15:43:33.427420Z	   41 Query	start transaction
2017-09-06T15:43:33.521411Z	   41 Query	INSERT INTO `import_customer` (`customer_id`, `import_id`, `name`)
VALUES (1, 5, 'cust');
2017-09-06T15:43:33.521411Z	   41 Query	SELECT `id` FROM `import_customer` WHERE `id` = LAST_INSERT_ID();
2017-09-06T15:43:33.522411Z	   41 Query	INSERT INTO `import_group` (`group_id`, `import_id`, `name`)
VALUES (1, 5, 'cat1');
2017-09-06T15:43:33.522411Z	   41 Query	SELECT `id` FROM `import_group` WHERE `id` = LAST_INSERT_ID();
2017-09-06T15:43:33.522411Z	   41 Query	INSERT INTO `import_group` (`group_id`, `import_id`, `name`)
VALUES (2, 5, 'cat2');
2017-09-06T15:43:33.522411Z	   41 Query	SELECT `id` FROM `import_group` WHERE `id` = LAST_INSERT_ID();
2017-09-06T15:43:33.522411Z	   41 Query	INSERT INTO `import_summary` (`import_id`, `value`)
VALUES (5, true);
2017-09-06T15:43:33.573405Z	   41 Query	SELECT ROW_COUNT();
2017-09-06T15:43:33.574405Z	   41 Query	INSERT INTO `import_zone` (`import_id`, `name`, `zone_id`)
VALUES (5, 'zone1', 1);
[...]

PR #407, opened 2017-10-17
Fix batch insert when some entities have no store columns are generated

neoGeneva:
This is to fix the issue I raised #405, and in turn #397 / #379


caleblloyd:
Thanks for the contribution! I'll publish a prerelease on nuget tomorrow.

PR #431, opened 2017-11-02
Added exponential backoff based on EF Core ExecutionStrategy code.

xcellsoft:
Exponential Backoff is now added to the MySqlRetryNoDependendiciesExecutionStrategy class that supports the MySqlConnectionSettings.GetSettings method.

The code for this was copied from the ExecutionStrategy class in EF Core.

Take note that the maxRetryDelay in EF Core is not the cumulative maximum amount of time for retry operations as a whole. It is actually the maximum amount of delay time between attempts and as the exponential backoff grows it can hit this limit and then that would be the maximum amount of time between retries.

For Example, I have maxRetryCount=15 and maxRetryDelay=5 seconds, here is how that plays out:

Attempt: 1, fail then  0 second delay (the formula always does an immediate attempt with no delay)
Attempt: 2, fail then  1.025 second delay
Attempt: 3, fail then  3.29 second delay
Attempt: 4, fail then  5 second delay
Attempt: 5, fail then  5 second delay
Attempt: 6, fail then  5 second delay
Attempt: 7, fail then  5 second delay

...and so on and so forth, once the formula reaches the maxRetryDelay it uses that exclusively moving forward.

Thanks!
-Chris

### DIFF
--- a/src/EFCore.MySql/Internal/MySqlRetryNoDependendiciesExecutionStrategy.cs
+++ b/src/EFCore.MySql/Internal/MySqlRetryNoDependendiciesExecutionStrategy.cs
@@ -36,7 +36,7 @@ namespace EFCore.MySql.Internal
         /// <summary>
         ///     A pseudo-random number generator that can be used to vary the delay between retries.
         /// </summary>
-        protected virtual Random Random { get; } = new Random();
+        private static Random Random { get; } = new Random();
 
 
         /// <summary>
--- a/src/EFCore.MySql/Internal/MySqlRetryNoDependendiciesExecutionStrategy.cs
+++ b/src/EFCore.MySql/Internal/MySqlRetryNoDependendiciesExecutionStrategy.cs
@@ -1,7 +1,10 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Internal;
+using Microsoft.EntityFrameworkCore.Storage;
 
 namespace EFCore.MySql.Internal
 {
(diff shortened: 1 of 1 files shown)


src/EFCore.MySql/Internal/MySqlRetryNoDependendiciesExecutionStrategy.cs
```diff
@@ -16,6 +19,27 @@ public class MySqlRetryNoDependendiciesExecutionStrategy
...
+
+        /// <summary>
+        ///     The default coefficient for the exponential function used to compute the delay between retries, must be nonnegative.
+        /// </summary>
+        private static readonly TimeSpan _defaultCoefficient = TimeSpan.FromSeconds(1);
+
+        /// <summary>
+        ///     A pseudo-random number generator that can be used to vary the delay between retries.
+        /// </summary>
+        protected virtual Random Random { get; } = new Random();
```
Kagamine: Should this be static member better? Or is this class `MySqlRetryNoDependendiciesExecutionStrategy` singleton?

xcellsoft: It’s not dependency injected class, it could just be private. I just copied verbatim from executionstrategy to keep the same pattern. Lmk if you want me to convert.

Good stuff though.

xcellsoft: It’s not dependency injected class, it could just be private. I just copied verbatim from executionstrategy to keep the same pattern. Lmk if you want me to convert.

Good stuff though.

Kagamine: I don't think generate a new Random instance each time will keep the random seed, if you can, please make the random object static.

xcellsoft: Done.

PR #435, opened 2017-11-06
Async Logging in MySqlRelationalCommand

caleblloyd:


caleblloyd:
Looks like [https://api.nuget.org/v3/index.json](https://api.nuget.org/v3/index.json) is returning a 504 right now, CI will fail until that is fixed.

PR #456, opened 2017-12-06
Don't cache exceptions in Lazy Initialization of MySqlConnectionSettings

### LINKED ISSUE #434: Problems connecting to MySQL with version 2.0
I use dependency injection to resolve DbContext objects and other services in non asp.net core environment. This works fine with the provider prior to version 2.0. After update to the latest 2.0 version the system turns unstable and I get randoms exceptions of 'Unable to connect to any of the specified MySQL hosts'.

The following construction that I used successfuly prior to 2.0 is working with problems after update.

`dbContext.Database.GetDbConnection()`

The first time I try to get a dbconnection this way it throws 'Unable to connect to any of the specified MySQL hosts' or it tries to connect inmediately and fails because the connection string is not properly configured. The workaround here is to catch the exception and retry the operation, the second time instead of return a null dbconnection it returns a valid connection that I can configure and properly use. Prior to 2.0 the same code worked flawlessly.

Some other normal usage of DbContext connections are also failing related of how I use the scope to get the services from dependency injection. The key here is that this doesn't  occur with previous version. In general I am just reporting this strange behavior with connections in my case.

OS: Win10
MySQL: 8.x
Pomelo MySQL: 2.0.0.1

caleblloyd:
- Fixes #434

### DIFF
--- a/src/EFCore.MySql/Internal/MySqlOptions.cs
+++ b/src/EFCore.MySql/Internal/MySqlOptions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Data;
 using System.Data.Common;
+using System.Threading;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Infrastructure.Internal;
 using Microsoft.EntityFrameworkCore.Storage;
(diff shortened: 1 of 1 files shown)


caleblloyd:
@yukozh can you bump the MyGet version for the nightly build?

yukozh:
all right

PR #458, opened 2017-12-20
Select for Update branch

conradmicallef:
Good morning, I would like you to consider pulling this branch as it gives options to control locking using Select ... FOR UPDATE which in transactional cases resolves a lot of headaches as it avoids the Concurrent Optimistic Exceptions


caleblloyd:
This is an all-or-nothing approach - if enabled it runs all queries as `SELECT FOR UPDATE`

`SELECT FOR UPDATE` is typically wrapped in a transaction and followed by an `UPDATE`.  It gives the transaction exclusive access to the selected rows.  Blanket enabling it will cause a lot of extra locking/unlocking for simple queries that do not wrap it in transactions.

The approach needs to be re-worked so that `SELECT FOR UPDATE` is enabled on a per-query basis, not an all-or-nothing approach.  I'm unsure if there is an easy way to do that in Entity Framework.  It may be best just to use [Raw SQL Queries](https://docs.microsoft.com/en-us/ef/core/querying/raw-sql) in parts of your application that need `SELECT FOR UPDATE` functionality.

PR #500, opened 2018-03-02
Fixes #497, Upgrade to EF Core 2.1.0-preview1-final

### LINKED ISSUE #497: MySQL provider is using internal code
See issue reported here: https://github.com/aspnet/EntityFrameworkCore/issues/11078

RelationalParameterBuilder is internal code, which changes from release to release. This makes using internal code in a provider a very bad idea because it breaks anyone using that provider when the internal code is changed. This is preventing MySQL from being used with EF Core 2.1.

The MySql provider should be updated to use no internal EF code. If there are places where using such code is needed, then these should be communicated to the EF team such that we can address the issue or suggest workarounds.

yukozh:


yukozh:
Hi @caleblloyd, Could you help to upgrade the build script?

caleblloyd:
@yukozh I don't know what the problem is with AppVeyor, it is acting up

For this PR, we need to continue to target `netcoreapp2.0` and use `Microsoft.AspNetCore.All=2.0.0`.  The reason is EF `2.1.0-preview1-final` works fine with `netcoreapp2.0` and it is too much of a pain to chase down preview releases of `netcoreapp2.1`

I made those changes in this PR in 84a91a5f5f237f88f35abc8dac72da98ef5febef

The build is broken, `dotnet build` throws some errors and warnings.  Some updates to our codebase need to be made for compatibility with EF `2.1`, I can work on that throughout the coming weeks.

caleblloyd:
Going to close this and work on a 2.1-preview upgrade branch.

PR #501, opened 2018-03-02
Fixes #499

### LINKED ISSUE #499: MySQL column type "date" not being mapped
### Steps to reproduce
Ideally include a complete code listing that we can run to reproduce the issue.
Alternatively, you can provide a project/solution that we can run.

### The issue
Running `Scaffold-DbContext -Connection "[connection string]" -Provider "Pomelo.EntityFrameworkCore.MySql" -OutputDir "Models\Cms" -Context "CmsContext"` causes any mysql column of the `date` type to not get mapped.

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Exception message: Could not find type mapping for column 'tblname.colname' with data type 'date'. Skipping column.
Stack trace:
```

### Further technical details

MySQL version: 
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 2.0.1

Could the MySQL date type not be mapped to a C# DateTime object? If you can clue me in to where these mappings are I don't mind trying to help on this issue.

JaredAtAdevsco:
I have used this locally to solve quite a few errors that were caused while trying to generate the database contexts for a project. 

Normally I would just use the DateTime format on the MySQL box, but I don't have access to update the schema.

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/MySqlDateTypeMapping.cs
+++ b/src/EFCore.MySql/Storage/Internal/MySqlDateTypeMapping.cs
@@ -1,0 +1,54 @@
+// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System;
+using System.Data;
+using JetBrains.Annotations;
+
+//ReSharper disable once CheckNamespace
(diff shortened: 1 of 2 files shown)


caleblloyd:
Thanks for the contribution, @JaredAtAdevsco!

Issue #393, opened 2017-10-04
Rename column operation being ran against the wrong database when using multiple DbContexts


robert396:
### Steps to reproduce
[source example](https://github.com/robert396/multipledbcontextissue)

### The issue
When using multiple DbContext's targeting different databases, the RenameColumn operation is executed using the name of the first database as the schema which causes the `GetCreateTable` method of the `MySqlMigrationsSqlGenerator` to fail when migrating via code.

First DbContext: `ClientDbContext` -> `sample_clients` 
Second DbContext: `SiteDbContext` -> `sample_sites`
Table column to rename: `SiteDomain` to `Domain`
Method ran: `siteDbContext.Database.Migrate()`

```
Exception message: Table 'sample_clients.sites' doesn't exist
Stack trace:
   at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet(ResultSet resultSet)
   at MySql.Data.MySqlClient.MySqlDataReader.<ReadFirstResultSetAsync>d__64.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.MySqlDataReader.<CreateAsync>d__63.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.CommandExecutors.TextCommandExecutor.<ExecuteReaderAsync>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.MySqlCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at Microsoft.EntityFrameworkCore.Internal.MySqlOptions.ExecuteCreateTable(DbConnection connection, ISqlGenerationHelper sqlGenerationHelper, String table, String schema)
[...]

PR #538, opened 2018-05-09
Rewrite type mapping for the 2.1 API

### LINKED ISSUE #512: 2.1.0-preview1 Upgrade 'GeneratedTime.CreatedTimestamp' could not be mapped
Working on the update to 2.1.0-preview1:

- Branch is [here](https://github.com/caleblloyd/Pomelo.EntityFrameworkCore.MySql/tree/f_2.1.0-preview)
- PR is https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/510

Presently everything is failing with this error and I can't figure out what is missing:

```
[xUnit.net 00:00:03.3297454]       System.InvalidOperationException : The property 'GeneratedTime.CreatedTimestamp' could not be mapped, because it is of type 'DateTime' which is not a supported primitive type or a valid entity type. Either explicitly map this property, or ignore it using the '[NotMapped]' attribute or by using 'EntityTypeBuilder.Ignore' in 'OnModelCreating'.
[xUnit.net 00:00:03.3298187]       Stack Trace:
[xUnit.net 00:00:03.3298596]            at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.PropertyMappingValidationConvention.Apply(InternalModelBuilder modelBuilder)
[xUnit.net 00:00:03.3298985]            at Microsoft.EntityFrameworkCore.Metadata.Conventions.Internal.ConventionDispatcher.ImmediateConventionScope.OnModelBuilt(InternalModelBuilder modelBuilder)
[xUnit.net 00:00:03.3299433]            at Microsoft.EntityFrameworkCore.Infrastructure.ModelSource.CreateModel(DbContext context, IConventionSetBuilder conventionSetBuilder, IModelValidator validator)
[xUnit.net 00:00:03.3299740]            at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
[xUnit.net 00:00:03.3299972]            at Microsoft.EntityFrameworkCore.Internal.DbContextServices.CreateModel()
[xUnit.net 00:00:03.3300151]            at Microsoft.EntityFrameworkCore.Internal.DbContextServices.get_Model()
[xUnit.net 00:00:03.3300331]            at lambda_method(Closure , ServiceProviderEngineScope )
[xUnit.net 00:00:03.3300568]            at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
[xUnit.net 00:00:03.3300821]            at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
[xUnit.net 00:00:03.3301011]            at Microsoft.EntityFrameworkCore.DbContext.get_DbContextDependencies()
[xUnit.net 00:00:03.3301269]            at Microsoft.EntityFrameworkCore.DbContext.get_InternalServiceProvider()
[xUnit.net 00:00:03.3301454]            at Microsoft.EntityFrameworkCore.DbContext.get_DbContextDependencies()
[xUnit.net 00:00:03.3301660]            at Microsoft.EntityFrameworkCore.DbContext.EntryWithoutDetectChanges[TEntity](TEntity entity)
[xUnit.net 00:00:03.3301856]            at Microsoft.EntityFrameworkCore.DbContext.SetEntityState[TEntity](TEntity entity, EntityState entityState)
[xUnit.net 00:00:03.3302140]         /home/caleb/Projects/caleb/Pomelo.EntityFrameworkCore.Mysql/test/EFCore.MySql.FunctionalTests/Tests/Models/ExpressionTest.cs(32,0): at EFCore.MySql.FunctionalTests.Tests.Models.ExpressionTest..ctor()

```

Opened upstream at: https://github.com/aspnet/EntityFrameworkCore/issues/11233, received this response:

> That exception means that the the type mapper (ITypeMappingSource in 2.1) is not returning a mapping for DateTime. ITypeMappingSource calls into your implementation of IRelationalTypeMapper if you haven't implemented the new service shapes directly.

> I would suggest debugging to see why the type mapper is not returning a mapping on this code path.

Looking for help from someone who can step-through debug this issue and find the root cause.  Otherwise I can work on it as I get time, but that could delay a 2.1.0-preview1 release as I don't have a lot of free time in the next couple of weeks

AndriySvyryd:
Fixes #512

This fixes all the old tests. There are still some upstream tests failing, mainly due to #393 and some issues in query SQL generation.


src/EFCore.MySql/Storage/Internal/ByteArrayComparer.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System.Collections;
+using System.Linq;
+using Microsoft.EntityFrameworkCore.ChangeTracking;
+
+namespace EFCore.MySql.Storage.Internal
+{
+    public class ByteArrayComparer : ValueComparer<byte[]>
```
ajcvickers: Why is this needed?

AndriySvyryd: To compare the binary representations of `timestamp`. Is there a better way of doing this?

src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
```diff
@@ -0,0 +1,378 @@
...
+            new MySqlDateTimeTypeMapping("timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP");
+        private readonly RelationalTypeMapping _rowversion6 =
+            new MySqlDateTimeTypeMapping("timestamp(6) DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP");
+
+        private readonly RelationalTypeMapping _binaryRowversion =
+            new MySqlDateTimeTypeMapping("timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP",
+                new BytesToDateTimeConverter(), new ByteArrayComparer());
+        private readonly RelationalTypeMapping _binaryRowversion6 =
+            new MySqlDateTimeTypeMapping("timestamp(6) DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP",
+                new BytesToDateTimeConverter(), new ByteArrayComparer());
```
AndriySvyryd: @ajcvickers usages of `ByteArrayComparer`

src/EFCore.MySql/Storage/Internal/ByteArrayComparer.cs
ajcvickers: Just create a ValueComparer instance passing in "favorStructuralComparisons" as true.

ajcvickers: Never mind. I see now that you need to snapshot it as well.

src/EFCore.MySql/Storage/Internal/MySqlJsonTypeMapping.cs
```diff
@@ -20,34 +22,75 @@ namespace EFCore.MySql.Storage.Internal
...
+                new RelationalTypeMappingParameters(
+                    new CoreTypeMappingParameters(clrType, GetConverter(clrType), GetComprarer(clrType)),
+                    storeType ?? "json",
+                    StoreTypePostfix.None,
+                    System.Data.DbType.String,
+                    unicode ?? false))
+        {
+        }
+
+        private static ValueConverter GetConverter(Type jsonType)
```
ajcvickers: Do these need to be thread-safe?

AndriySvyryd: At worst it would just create the converter twice.

ajcvickers: But TryGetValue and insertion into the Dictionary is not thread-safe, so if two threads try to do this at the same time it would not be safe--use a ConcurrentDictionary instead?

PR #539, opened 2018-05-12
Provide MySQL syntax overrides for a number of spec test classes

csboling:
In search of low-hanging fruit, I noticed that a number of tests were failing due to string comparisons with generated SQL which used double quotes or square brackets rather than MySQL-style backticks. This included some tests for stored procedures, which didn't appear to exist in the test schema, so I also modified MySqlTestStore to define the Northwind stored procedures, and got the `FromSqlSproc`-type tests to pass. There might be a better place to execute that SQL but this seemed to work well enough. A number of other failures were coming from expecting SQL functions in lower case, while the SQL generator was emitting them capitalized, and things like that. Some were also coming from multiline statements which were using fixed \n linebreaks instead of dynamically choosing the right "EOL" string.

All together I think these changes got something like 65 more tests to pass. I think there are a number of other tests failing due to issues in this vein, especially related to double-quote vs. backtick escaping of table/column names, because some tests check generated SQL strings in other ways that Assert.Equal and it wasn't as clear to me how to make sure backticks were used. There are also some other MySQL syntax errors still whose causes were not as clear to me.


test/EFCore.MySql.UpstreamFunctionalTests/TestUtilities/MySqlTestStore.cs
```diff
@@ -67,6 +67,25 @@ protected override void Initialize(Func<DbContext> createContext, Action<DbConte
...
+CREATE PROCEDURE `CustOrderHist` (IN CustomerID VARCHAR(768))
+BEGIN
+  SELECT `ProductName`, SUM(`Quantity`) AS `Total`
+  FROM `Products` `p`, `Order Details` `od`, `Orders` `o`, `Customers` `c`
+  WHERE `c`.`CustomerId` = `CustomerId`
+  AND `c`.`CustomerId` = `o`.`CustomerId`
+  AND `o`.`OrderId` = `od`.`OrderId`
+  AND `od`.`ProductId` = `p`.`ProductId`
+  GROUP BY `ProductName`;
+END;");
```
AndriySvyryd: This should be done in `NorthwindQueryMySqlFixture.Seed`

test/EFCore.MySql.UpstreamFunctionalTests/MigrationSqlGeneratorMySqlTest.cs
```diff
@@ -514,7 +514,7 @@ INTO PRIMARY_KEY_COLUMN_NAME
 		EXECUTE SQL_EXP_EXECUTE;
 		DEALLOCATE PREPARE SQL_EXP_EXECUTE;
 	END IF;
-END;" + EOL +
+END;".Split('\n')) + EOL +
```
AndriySvyryd: To make this resilient to future line ending changes it would be better to call:
```
.Replace("\r", string.Empty).Replace("\n", EOL)
```

PR #548, opened 2018-05-22
Map ENUM and SET to strings

fmauNeko:
Hello,

Here's a quickfix to map enums and sets to strings.

Have a good day !


caleblloyd:
Thanks for the PR.  Type Mapping is changing in 2.1.0 and all of that work is happening in the [f_2.1.0-spec-tests](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/tree/f_2.1.0-spec-tests)

There are additional tests that would be needed to verify this type mapping.  I'm going to close this PR, feel free to re-open one against `f_2.1.0-spec-tests` or open an issue to track the feature request.

PR #569, opened 2018-05-30
Use Default Charset

### LINKED ISSUE #553: MySQL collation issue in EF Core
My database is MySQL. I generate migration and update database, then I see the collation of table which is different on different machines. On my machine its latin1 - default collation and on a different machine, it is utf8 - default collation.

So due to this behavior, I'm getting an exception while inserting Japanese word (日本語) into a table which has a varchar(64) column. It works fine on another machine which has utf8 - default collation on this table.

EF Core Version: 2.0.1
MySql: 5.7.22
Pomelo.EntityFrameworkCore.MySql: 2.0.1

caleblloyd:
- Fixes #553 
- Uses the Database Default Charset for both Unicode and ANSI strings

### DIFF
--- a/src/EFCore.MySql/Infrastructure/CharSetAppendBehavior.cs
+++ b/src/EFCore.MySql/Infrastructure/CharSetAppendBehavior.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure
+{
+    [Flags]
+    public enum CharSetBehavior
+    {
+        NeverAppend = 0,
+        AppendToAnsiIndexAndKeyColumns       = 1 << 0,
+        AppendToUnicodeIndexAndKeyColumns    = 1 << 1,
+        AppendToAnsiNonIndexAndKeyColumns    = 1 << 2,
+        AppendToUnicodeNonIndexAndKeyColumns = 1 << 3,
+        AppendToAllKeyAndIndexColumns = AppendToAnsiIndexAndKeyColumns | AppendToUnicodeIndexAndKeyColumns,
+        AppendToAllAnsiColumns = AppendToAnsiIndexAndKeyColumns | AppendToAnsiNonIndexAndKeyColumns,
+        AppendToAllUnicodeColumns = AppendToUnicodeIndexAndKeyColumns | AppendToUnicodeNonIndexAndKeyColumns,
+        AppendToAllColumns = AppendToAllAnsiColumns | AppendToAllUnicodeColumns
+    }
+}
--- a/src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
+++ b/src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
@@ -23,6 +23,9 @@ namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal
             : base(copyFrom)
         {
             ServerVersion = copyFrom.ServerVersion;
+            NullableCharSetBehavior = copyFrom.NullableCharSetBehavior;
+            AnsiCharSetInfo = copyFrom.AnsiCharSetInfo;
+            UnicodeCharSetInfo = copyFrom.UnicodeCharSetInfo;
         }
 
         protected override RelationalOptionsExtension Clone()
@@ -33,6 +36,26 @@ namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
         public ServerVersion ServerVersion { get; private set; }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public CharSetBehavior? NullableCharSetBehavior { get; private set; }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public CharSetInfo AnsiCharSetInfo { get; private set; }
+
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public CharSetInfo UnicodeCharSetInfo { get; private set; }
+
 
         /// <summary>
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
@@ -51,11 +74,54 @@ namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal
         ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
         ///     directly from your code. This API may change or be removed in future releases.
         /// </summary>
+        public MySqlOptionsExtension WithCharSetBehavior(CharSetBehavior charSetBehavior)
+        {
+            var clone = (MySqlOptionsExtension)Clone();
+
+            clone.NullableCharSetBehavior = charSetBehavior;
+
+            return clone;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public MySqlOptionsExtension WithAnsiCharSetInfo(CharSetInfo charSetInfo)
+        {
+            var clone = (MySqlOptionsExtension)Clone();
+
+            clone.AnsiCharSetInfo = charSetInfo;
+
+            return clone;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
+        public MySqlOptionsExtension WithUnicodeCharSetInfo(CharSetInfo charSetInfo)
+        {
+            var clone = (MySqlOptionsExtension)Clone();
+
+            clone.UnicodeCharSetInfo = charSetInfo;
+
+            return clone;
+        }
+
+        /// <summary>
+        ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+        ///     directly from your code. This API may change or be removed in future releases.
+        /// </summary>
         public override long GetServiceProviderHashCode()
         {
             if (_serviceProviderHash == null)
             {
-                _serviceProviderHash = (base.GetServiceProviderHashCode() * 397) ^ (ServerVersion?.GetHashCode() ?? 0L);
+                _serviceProviderHash = (base.GetServiceProviderHashCode() * 397) ^
+                                       (ServerVersion?.GetHashCode() ?? 0L) ^
+                                       (NullableCharSetBehavior?.GetHashCode() ?? 0L) ^
+                                       (AnsiCharSetInfo?.GetHashCode() ?? 0L) ^
+                                       (UnicodeCharSetInfo?.GetHashCode() ?? 0L);
             }
 
             return _serviceProviderHash.Value;
--- a/src/EFCore.MySql/Infrastructure/MySqlDbContextOptionsBuilder.cs
+++ b/src/EFCore.MySql/Infrastructure/MySqlDbContextOptionsBuilder.cs
@@ -32,6 +32,24 @@ namespace Microsoft.EntityFrameworkCore.Infrastructure
             => WithOption(e => e.WithServerVersion(new ServerVersion(serverVersion)));
 
         /// <summary>
+        ///     Configures the Default CharSet Behavior
+        /// </summary>
+        public virtual MySqlDbContextOptionsBuilder CharSetBehavior(CharSetBehavior charSetBehavior)
+            => WithOption(e => e.WithCharSetBehavior(charSetBehavior));
+
+        /// <summary>
+        ///     Configures the ANSI CharSet
+        /// </summary>
+        public virtual MySqlDbContextOptionsBuilder AnsiCharSet(CharSet charSet)
+            => WithOption(e => e.WithAnsiCharSetInfo(new CharSetInfo(charSet)));
+
+        /// <summary>
+        ///     Configures the Unicode CharSet
+        /// </summary>
+        public virtual MySqlDbContextOptionsBuilder UnicodeCharSet(CharSet charSet)
+            => WithOption(e => e.WithUnicodeCharSetInfo(new CharSetInfo(charSet)));
+
+        /// <summary>
         ///     Configures the context to use the default retrying <see cref="IExecutionStrategy" />.
         /// </summary>
         public virtual MySqlDbContextOptionsBuilder EnableRetryOnFailure()
--- a/src/EFCore.MySql/Internal/MySqlOptions.cs
+++ b/src/EFCore.MySql/Internal/MySqlOptions.cs
@@ -7,6 +7,7 @@ using Pomelo.EntityFrameworkCore.MySql.Storage.Internal;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Internal;
+using Pomelo.EntityFrameworkCore.MySql.Infrastructure;
 
 namespace Pomelo.EntityFrameworkCore.MySql.Internal
 {
@@ -16,6 +17,9 @@ namespace Pomelo.EntityFrameworkCore.MySql.Internal
         {
             ConnectionSettings = new MySqlConnectionSettings();
             ServerVersion = new ServerVersion(null);
+            CharSetBehavior = CharSetBehavior.AppendToAllAnsiColumns;
+            AnsiCharSetInfo = new CharSetInfo(CharSet.Latin1);
+            UnicodeCharSetInfo = new CharSetInfo(CharSet.Utf8mb4);
         }
 
         public virtual void Initialize(IDbContextOptions options)
@@ -24,6 +28,9 @@ namespace Pomelo.EntityFrameworkCore.MySql.Internal
 
             ConnectionSettings = GetConnectionSettings(mySqlOptions);
             ServerVersion = mySqlOptions.ServerVersion ?? ServerVersion;
+            CharSetBehavior = mySqlOptions.NullableCharSetBehavior ?? CharSetBehavior;
+            AnsiCharSetInfo = mySqlOptions.AnsiCharSetInfo ?? AnsiCharSetInfo;
+            UnicodeCharSetInfo = mySqlOptions.UnicodeCharSetInfo ?? UnicodeCharSetInfo;
         }
 
         public virtual void Validate(IDbContextOptions options)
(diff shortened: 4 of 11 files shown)


caleblloyd:
First round of Design Review ready.  Here's some reasoning:

- 4 Different Options:
  - **ServerDefaultCharSetInfo**: If this is set, it will be used to calculate `MaxSize` of a string
  - **AnsiCharSetInfo**: If this is set, it will add `CHARACTER SET` suffix for ANSI Strings
  - **UnicodeCharSetInfo**: If this is set, it will add `CHARACTER SET` suffix for Unicode Strings
  - **IndexAndKeysOnly**: Defaults to `false`.  If set to `true`, ANSI and Unicode suffixes will only be applied to Indexes and Keys

Opted for a Config Object that can be used with an Object Initializer instead of writing 4 `WithXXX` methods.  Open for changing if this is too far outside of the norm though.

src/EFCore.MySql/Infrastructure/CharSetConfig.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using Pomelo.EntityFrameworkCore.MySql.Storage.Internal;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure
+{
+    public class CharSetConfig
+    {
+        public CharSetInfo ServerDefaultCharSetInfo { get; private set; }
```
AndriySvyryd: Since we only need the bytes per character I think it would be less confusing to the user for this to be an `int`

src/EFCore.MySql/Infrastructure/CharSetConfig.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public CharSetInfo UnicodeCharSetInfo { get; private set; }
+
+        public string UnicodeCharSet
+        {
+            get => UnicodeCharSetInfo?.CharSetName;
+            set => UnicodeCharSetInfo = new CharSetInfo(value);
+        }
+
+        public bool IndexAndKeysOnly { get; set; } = false;
```
AndriySvyryd: Needs a more descriptive name, like `UseNonDefaultCharsetForIndexAndKeysOnly`

src/EFCore.MySql/Infrastructure/MySqlDbContextOptionsBuilder.cs
```diff
@@ -31,6 +31,12 @@ public virtual MySqlDbContextOptionsBuilder ServerVersion(Version version, Serve
         public virtual MySqlDbContextOptionsBuilder ServerVersion(string serverVersion)
             => WithOption(e => e.WithServerVersion(new ServerVersion(serverVersion)));
 
+        /// <summary>
+        ///     Configures the Default CharSet
+        /// </summary>
+        public virtual MySqlDbContextOptionsBuilder CharSetConfig(CharSetConfig charSetConfig)
```
AndriySvyryd: Instead of taking a mutable options object we prefer to present a Fluent API builder, see [WarningsConfigurationBuilder.cs](https://github.com/aspnet/EntityFrameworkCore/blob/5b894d9edf8aecd240629a4ee6149470353a65f1/src/EFCore/Diagnostics/WarningsConfigurationBuilder.cs) and [WarningsConfiguration.cs](https://github.com/aspnet/EntityFrameworkCore/blob/280f2ce4bcf267d09fed7e3f834ac1ac40966d9d/src/EFCore/Diagnostics/WarningsConfiguration.cs)

src/EFCore.MySql/Infrastructure/CharSetConfig.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public CharSetInfo AnsiCharSetInfo { get; private set; }
+
+        public string AnsiCharSet
+        {
+            get => AnsiCharSetInfo?.CharSetName;
+            set => AnsiCharSetInfo = new CharSetInfo(value);
+        }
+
+        public CharSetInfo UnicodeCharSetInfo { get; private set; }
```
AndriySvyryd: I think it would be better to have the charsets set to some value by default as it would make the model independent from the server. One less thing for users to worry about unless they opt-in to follow the server default.

caleblloyd:
Second Round of Design:

Based on feedback and looking at this again, I agree that it is best to simplify and pick some defaults.  Here is the new design which is pretty much https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/553#issuecomment-393305238

- **DefaultCharSetBehavior**:
  - None (default, do not append charset)
  - IndexAndKeyColumns (only append charset to Index and Key Columns)
  - AllColumns (append charset to all columns)
- **AnsiCharSetInfo**: Defaults to `utf8mb4`
- **UnicodeCharSetInfo**: Defaults to `utf8mb4`

The reason to default both to `utf8mb4` is because:
1. It is what we recommend in README.md
2. It is safe because it is the largest `BytesPerChar` out of the `CharSet` that are included
3. Lots of people set this to the default to support different languages
4. It is the default in MySQL 8

AndriySvyryd:
@caleblloyd Note that by default the users will get Unicode columns. They would need to call `.IsUnicode(false)` on a property to get the ANSI mapping.
So considering that if the user doesn't configure anything they would get `utf8mb4` I think the default for ANSI should be `latin1` because it would be confusing that calling `.IsUnicode(false)` has no effect.

caleblloyd:
> Note that by default the users will get Unicode columns. They would need to call .IsUnicode(false) on a property to get the ANSI mapping.

Yes, in that case it does make sense to default ANSI to `Latin1`.  I have made that change.

src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
```diff
@@ -329,13 +330,27 @@ private RelationalTypeMapping FindRawMapping(RelationalTypeMappingInfo mappingIn
...
-                    var suffix = isAnsi ? " CHARACTER SET latin1" : " CHARACTER SET ucs2";
-                    var maxSize = isAnsi ? 8000 : 4000;
+                    var charSetSuffix = "";
+                    var bytesPerChar = isAnsi
+                        ? _options.AnsiCharSetInfo.BytesPerChar
+                        : _options.UnicodeCharSetInfo.BytesPerChar;
+
+                    if (_options.DefaultCharSetBehavior == DefaultCharSetBehavior.AllColumns ||
+                        (_options.DefaultCharSetBehavior == DefaultCharSetBehavior.IndexAndKeyColumns &&
+                         mappingInfo.IsKeyOrIndex))
```
AndriySvyryd: This is backwards. If `DefaultCharSetBehavior` is `DefaultCharSetBehavior.AllColumns` then no charset should be added so the column use the server default.

AndriySvyryd: You could keep this behavior, but rename `DefaultCharSetBehavior`

Giantaxe:
@caleblloyd 
Just noticed your issue about renaming indexes at MySQL 5.5 [https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/576]Issue #576(url)

That reminded me, for your information. UTF8MB4 was added around MySQL 5.5.3. In case people do use an older version then that there should be a fallback character set. 

See: [http://ftp.nchu.edu.tw/MySQL/doc/refman/5.4/en/charset-unicode.html]http://ftp.nchu.edu.tw/MySQL/doc/refman/5.4/en/charset-unicode.html(url)

That is if you do want to support anything older then 5.5.3 to begin with.

src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
caleblloyd: I think the options need to be clarified and expanded:

- CharSetBehavior.AppendToAllAnsiColumns (default)
- CharSetBehavior.AppendToAllUnicodeColumns
- CharSetBehavior.AppendToAllColumns
- CharSetBehavior.AppendToAnsiIndexAndKeyColumns
- CharSetBehavior.AppendToUnicodeIndexAndKeyColumns
- CharSetBehavior.AppendToIndexAndKeyColumns
- CharSetBehavior.NeverAppend

caleblloyd:
@Giantaxe if we use `CharSetBehavior.AppendToAllAnsiColumns` as the default, it will not append `UTF8MB4` to anything by default.  I think not appending charset to Unicode by default is the right default.

If someone is using an older version of MySQL then they can explicitly set their unicode charset when they configure the `CharSetBehavior`

src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
AndriySvyryd: @caleblloyd Perhaps would be better to split this in two or make it a Flags enum because I think that a more useful default would be `AppendToAllAnsiColumns` and `AppendToUnicodeIndexAndKeyColumns` combined

caleblloyd: +1 for flags

Giantaxe: @AndriySvyryd Using flags, something like this?

My idea (I just thoughtthis up, can't really test it at the moment.)

```csharp
[Flags]
enum CharSetAppendBehavior
{
    Never,
    Always,
    KeyAndIndex
    Unicode,
    UnicodeKeyAndIndex
    Ansi,
    AnsiKeyAndIndex
}
```

If Unicode is the default charset, then something like this?
`DefaultCharSetBehavior = CharSetAppendBehavior.Ansi | CharSetAppendBehavior.AnsiKeyAndIndex;`

The result would be something like this.
```csharp
if (!_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.Never) && 
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.Always) ||
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.Ansi) && isAnsi) ||
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.Unicode) && !isAnsi) ||
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.KeyAndIndex) && mappingInfo.IsKeyOrIndex) ||
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.AnsiKeyAndIndex) && mappingInfo.IsKeyOrIndex && isAnsi) ||
    (_options.DefaultCharSetBehavior.HasFlag(CharSetAppendBehavior.UnicodeKeyAndIndex) && mappingInfo.IsKeyOrIndex && !isAnsi)))
{
	var charSetName = isAnsi
		? _options.AnsiCharSetInfo.CharSetName
		: _options.UnicodeCharSetInfo.CharSetName;
	charSetSuffix = $" CHARACTER SET {charSetName}";
}
```

That should do all the cases right?

AndriySvyryd: @Giantaxe I was thinking something more like this

``` C#
[Flags]
public enum CharSetAppendBehavior
{
    NeverAppend = 0,
    AppendToAnsiIndexAndKeyColumns       = 1 << 0,
    AppendToUnicodeIndexAndKeyColumns    = 1 << 1,
    AppendToAnsiNonIndexAndKeyColumns    = 1 << 2,
    AppendToUnicodeNonIndexAndKeyColumns = 1 << 3,
    AppendToAllKeyAndIndexColumns = AppendToAnsiIndexAndKeyColumns | AppendToUnicodeIndexAndKeyColumns,
    AppendToAllAnsiColumns = AppendToAnsiIndexAndKeyColumns | AppendToAnsiNonIndexAndKeyColumns,
    AppendToAllUnicodeColumns = AppendToUnicodeIndexAndKeyColumns | AppendToUnicodeNonIndexAndKeyColumns,
    AppendToAllColumns = AppendToAllAnsiColumns | AppendToAllUnicodeColumns
}
```

Giantaxe: @AndriySvyryd 
Truth be told, I did not know it was even possible to combine flags within the definition of the same enum.

Guess I just learned something new.

That looks to be the safest bet.

Though for the naming convention, it reads as follow when written.
For instance: `CharSetAppendBehavior.AppendToUnicodeNonIndexAndKeyColumns `
The word Append is now written 2 times in the enum + enum value.

Either set the enum name to `CharSetBehavior` as suggested caleblloyd or remove the `AppendTo` part from the values.

So either:
1.  `CharSetBehavior.AppendToUnicodeNonIndexAndKeyColumns `
or
2.  `CharSetAppendBehavior.UnicodeNonIndexAndKeyColumns `

caleblloyd:
Updated the design to use Flags from https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/569#discussion_r192518060

Going to work on tests next.

caleblloyd:
image

test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.cs
```diff
@@ -151,6 +153,45 @@ public virtual void CreateTableUlongAutoincrement()
...
+        [InlineData(false, false, CharSetAppendBehavior.AppendToAllUnicodeColumns, CharSet.Latin1, CharSet.Utf8mb4, "")]
+        [InlineData(false, false, CharSetAppendBehavior.AppendToAnsiIndexAndKeyColumns, CharSet.Latin1, CharSet.Utf8mb4, "")]
+        [InlineData(false, false, CharSetAppendBehavior.AppendToAllAnsiColumns, CharSet.Latin1, CharSet.Utf8mb4, "latin1")]
+        [InlineData(false, false, CharSetAppendBehavior.AppendToAllColumns, CharSet.Latin1, CharSet.Utf8mb4, "latin1")]
+        [InlineData(false, false, CharSetAppendBehavior.AppendToAllColumns, CharSet.Ucs2, CharSet.Utf8mb4, "ucs2")]
+        public virtual void AddColumnOperation_with_charset(bool isUnicode, bool isIndex, CharSetAppendBehavior charSetAppendBehavior,
+            CharSet ansiCharSet, CharSet unicodeCharSet, string expectedCharSet)
+        {
+            Generate(
+                modelBuilder => modelBuilder.Entity("Person").Property<string>("Name").IsUnicode(isUnicode),
```
caleblloyd: @AndriySvyryd how can I add `.IsIndex()` here so that the "Is Index" codepath can be tested?

AndriySvyryd: Something like
```C#
Generate(
    modelBuilder => modelBuilder.Entity(
        "Person",
        eb =>
        {
            eb.Property<string>("Name").IsUnicode(isUnicode);
            if (isIndex)
            {
                eb.HasIndex("Name");
            }
        }),
```

src/EFCore.MySql/Infrastructure/CharSetAppendBehavior.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure
+{
+    [Flags]
+    public enum CharSetAppendBehavior
```
AndriySvyryd: Did you consider dropping the `Append` as @Giantaxe suggested?

src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
```diff
@@ -55,7 +117,11 @@ public override long GetServiceProviderHashCode()
         {
             if (_serviceProviderHash == null)
             {
-                _serviceProviderHash = (base.GetServiceProviderHashCode() * 397) ^ (ServerVersion?.GetHashCode() ?? 0L);
+                _serviceProviderHash = (base.GetServiceProviderHashCode() * 397) ^
+                                       (ServerVersion?.GetHashCode() ?? 0L) ^
+                                       (NullableCharSetAppendBehavior?.GetHashCode() ?? 0L) ^
+                                       (AnsiCharSetInfo?.GetHashCode() ?? 0L) ^
+                                       (UnicodeCharSetInfo?.GetHashCode() ?? 0L);
```
AndriySvyryd: I'm pretty sure that none of these options need to be considered here as they don't affect which services go into the service provider, so you can just return `base.GetServiceProviderHashCode()`

src/EFCore.MySql/Internal/MySqlOptions.cs
```diff
@@ -16,6 +17,9 @@ public MySqlOptions()
         {
             ConnectionSettings = new MySqlConnectionSettings();
             ServerVersion = new ServerVersion(null);
+            CharSetAppendBehavior = CharSetAppendBehavior.AppendToAllAnsiColumns;
```
AndriySvyryd: Did you considered adding `AppendToUnicodeIndexAndKeyColumns` here? Does that make some of the functional tests fail?

caleblloyd: I think it's best to use the server default by default. If they have specified that it is ANSI, that should mean that Latin1 is safe. If they haven't specified anything, then it will default to Unicode. In that case I think it's best to use the server default, as users should pick a server default that is appropriate for the data they are storing.

src/EFCore.MySql/Infrastructure/CharSetAppendBehavior.cs
caleblloyd: Ok, I will simply call it `CharSetBehavior`

test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.cs
caleblloyd: Thanks I'll add that.

caleblloyd:
I'll make these changes later this evening. LGTM after these changes?

src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
caleblloyd: @AndriySvyryd these are necessary.  `MySqlOptions` is a singleton and these values get stored in `MySqlOptions`.  The tests failed with race conditions when I removed this.

src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
```diff
@@ -315,13 +316,39 @@ private RelationalTypeMapping FindRawMapping(RelationalTypeMappingInfo mappingIn
...
+                    }
+
+                    var maxSize = 8000 / bytesPerChar;
 
                     var size = mappingInfo.Size ??
                                (mappingInfo.IsKeyOrIndex
-                               // Allow to use at most half of the max key length, so at least 2 columns can fit
-                                   ? _options.ServerVersion.IndexMaxBytes / (isAnsi ? 2 : 4)
+                                   // Allow to use at most half of the max key length, so at least 2 columns can fit
+                                   ? Math.Min(_options.ServerVersion.IndexMaxBytes / (bytesPerChar * 2), 255)
```
caleblloyd: One noteworthy change - set the index to the minimum of half the available Index size, or 255.  This way 2 _or more_ columns can fit in the index.  255 has been our default in past releases, and there is no good reason to default to more.

PR #572, opened 2018-05-30
Fix the ComputedColumn convention

AndriySvyryd:
and some cleanup

@caleblloyd This doesn't fix everything, but should be enough to unblock you.


AndriySvyryd:
@caleblloyd The expected model depends on what the ValueGenerationStrategy on the model means. Currently if it's set to `Identity` than the properties that can are compatible will be marked with `Identity` and if it's set to `Computed` than the properties that can are compatible will be marked with `Computed`.
By convention it's set to `Identity`, so no columns are marked with `Computed`.

caleblloyd:
So what does an approach look like that allows both annotations?  That would be required for us to stay backwards-compatible with 1.x and 2.0.x migrations.

Basically all we need is:
- `ValueGenerated.OnAdd` => `MySqlValueGenerationStrategy.IdentityColumn` annotation
- `ValueGeneratedOnAddOrUpdate` => `MySqlValueGenerationStrategy.ComputedColumn` annotation

It seems odd to me that this is not the default behavior in SqlServer, the documentation seems to suggest it would be.  I feel like at some point this was the default behavior in SqlServer, did it change?

https://docs.microsoft.com/en-us/ef/core/modeling/generated-properties

AndriySvyryd:
@caleblloyd In that case you don't need the model annotation at all as there is no ambiguity.

The behavior has always been the same in SqlServer, though the implementation has changed. SqlServer does not provide any strategy for computed columns.

caleblloyd:
Thanks, I now understand the convention was used to switch between a "Identity" approach or a "Computed" approach.  I had incorrectly been under the impression that it was required to associate an annotation with `ValueGenerated.OnAdd/OnAddOrUpdated`

PR #575, opened 2018-05-31
more query fixes

maumar:
- added translation of IndexOf,
- fixed translation of DateAdd for more complex cases,
- skipped a bunch of tests that fail due to known issues.


maumar:
@AndriySvyryd

PR #579, opened 2018-06-01
Only consider rowversion for byte[] columns

### LINKED ISSUE #578: Only Allow `RowVersion` on `byte[]` Columns
Legacy migrations that used the `[Timestamp]` annotation on a `DateTime` column were never marked as `RowVersion` columns.  They were also never annotated with 

Example: 2.0.1 model:
```
	public class GeneratedConcurrencyToken
	{
		public int Id { get; set; }
		public int Gen { get; set; }
		[Timestamp]
		public DateTime Updated { get; set; }
	}
```

Corresponding migration:
```
            migrationBuilder.CreateTable(
                name: "GeneratedConcurrencyToken",
                columns: table => new
                {
                    Id = table.Column<int>(nullable: false)
                        .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn),
                    Gen = table.Column<int>(nullable: false),
                    Updated = table.Column<DateTime>(nullable: false)
                },
                constraints: table =>
                {
                    table.PrimaryKey("PK_GeneratedConcurrencyToken", x => x.Id);
                });
```

This causes a fresh migration up in 2.1.0 to fail, because `ON UPDATE CURRENT_TIMESTAMP` is never applied in the migration:
```
Field 'Updated' doesn't have a default value
```

The reason for this appears to be that `DateTime` columns have never been marked as `RowVersion` in the past.  `RowVersion` is explicitly reserved for `byte[]` columns in `EFCore.Relational`:

https://github.com/aspnet/EntityFrameworkCore/blob/2.1.0/src/EFCore.Relational/Migrations/Internal/MigrationsModelDiffer.cs#L962-L964

The best course of action here seems to be to fall in line with SQL Server and only allow `RowVersion` on `byte[]` Columns

If someone wants a concurrency check on a `DateTime` column, it should be done with:

```
	public class GeneratedConcurrencyToken
	{
		public int Id { get; set; }
		public int Gen { get; set; }
		[ConcurrencyCheck]
		[DatabaseGenerated(DatabaseGeneratedOption.Computed)]
		public DateTime Updated { get; set; }
	}
```

caleblloyd:
- Fixes #578

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   - echo "Tests" && ./dotnet-run.sh dotnet test test/EFCore.MySql.Tests
   # Too many Functional Tests failing; fix before re-enabling in CI
   # - echo "Functional Tests" && ./dotnet-run.sh dotnet test -c Release test/EFCore.MySql.FunctionalTests
-  - echo "Functional Migration Tests" && ./dotnet-run.sh dotnet test -c Release test/EFCore.MySql.FunctionalTests --filter "FullyQualifiedName~Pomelo.EntityFrameworkCore.MySql.FunctionalTests.MigrationSqlGeneratorMySqlTest|FullyQualifiedName~Pomelo.EntityFrameworkCore.MySql.FunctionalTests.MigrationSqlGeneratorMySql55Test"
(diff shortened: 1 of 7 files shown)


caleblloyd:
Well that's the first time CI has passed in a long time :smile:

caleblloyd:
Functional tests are looking better but still a decent amount of failures:

```
Total tests: 7299. Passed: 7106. Failed: 50. Skipped: 143.
```

I'll make an action item to open issues for the failures and add [Skip] with references to issues.  Once they are all documented in issues and skipped we can enable them to more easily detect regressions.

AndriySvyryd:
Ok, but do the optimistic concurrency tests pass for this change?

caleblloyd:
:+1: just enabled the optimistic concurrency test class in CI

Issue #461, opened 2017-12-28
Query: Invalid SQL generated for Guid if they are binary (OldGuids=true) when using "Contains()"


OrihuelaConde:
### Steps to reproduce
```
HashSet<Guid> productosSucursalId = _context.Set<ProductoSucursal>().Where(ps => ps.SucursalId == SucursalId).Select(ps => ps.ProductoSucursalId).ToHashSet();
.
.
.
entities = _context.Set<PedidoProductoSucursal>().Where(x => productosSucursalId.Contains(x.ProductoSucursalId)).ToArray();
```

### The issue
The Query generated by this is:

```
info: Microsoft.EntityFrameworkCore.Database.Command[20101]
      Executed DbCommand (0ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
      SELECT `x`.`PedidoId`, `x`.`ProductoSucursalId`, `x`.`CantidadDespachada`, `x`.`CantidadIngresada`, `x`.`CantidadPedida`, `x`.`Detalle`
      FROM `PedidoProdSuc` AS `x`
      WHERE `x`.`ProductoSucursalId` IN ('ef2d09e8-b3da-48d6-b6f8-eeb5720ccc79')
```

And should be "WHERE 'x'.'ProductoSucursalId' IN (0xE8092DEFDAB3D648B6F8EEB5720CCC79)".

### Further technical details

MySQL version: 5.7.17
Operating system: Windows 10 Pro 1709 (x64)
Pomelo.EntityFrameworkCore.MySql version: 2.0.1

Other details about my project setup:
Connection String: OldGuids=true

sico85:
Hello,
The same for me.

```
Executed DbCommand (1ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
`SELECT `x`.`ID`, `x`.`ModificationDate`, `x`.`Name` 
FROM `Products` AS `x`
WHERE `x`.`ID` = '00000001-0001-0001-0101-010101010101'
ORDER BY `x`.`ID`
LIMIT 1`
```

And should be WHERE `x`.`ID` = 0x01000000010001000101010101010101

I'm using MariaDB 10.1.30 with Pomelo.EntityFrameworkCore.MySql version: 2.0.0.1
Connection String: OldGuids=true

Thank you for your help.

AccessViolator:
@OrihuelaConde FYI Contains() works fine for me without OldGuids=true:
```
[...]

PR #584, opened 2018-06-02
Introduce custom type mapping for old guid literals.

robbieknuth:
The entity framework does not parameterize guids when they are put into a "IN" expression. They are generated as literals and put as a part of the SQL. This addresses issue #461 . Ultimately the fix is fairly straight forward: when using the old guid way, override the type mapping to provide an explicit implementation for the literal. I have run some preliminary tests:

Query looks like this:
``` c#
var records = await context.GuidsTest
    .Where(x => shouldExistListOfGuids.Contains(x.NotNullGuid))
    .ToListAsync();
```

schema is:

```
CREATE TABLE `GuidsTest` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `notNullGuid` binary(16) NOT NULL,
  `nullGuid` binary(16) DEFAULT NULL,
  PRIMARY KEY (`id`)
```

**Without** the fix, not oldguids:
```
      SELECT x.Id, x.NotNullGuid, x.NullGuid
      FROM GuidsTest AS x
      WHERE x.NotNullGuid IN ('01234567-1234-1234-1234-123456789012')
Record count: 1
```
Works as expcted. **With** the fix, not oldguids:
```
      SELECT x.Id, x.NotNullGuid, x.NullGuid
      FROM GuidsTest AS x
      WHERE x.NotNullGuid IN ('01234567-1234-1234-1234-123456789012')
Record count: 1
```
Works as expected. So this does not regress the non-oldguids case.

**Without** the fix, oldguids (This is the broken one):
```
      SELECT x.Id, x.NotNullGuid, x.NullGuid
      FROM GuidsTest AS x
      WHERE x.NotNullGuid IN ('01234567-1234-1234-1234-123456789012')
Record count: 0
```

And finally with the fix, old guids:
```
      SELECT x.Id, x.NotNullGuid, x.NullGuid
      FROM GuidsTest AS x
      WHERE x.NotNullGuid IN (0x67452301341234121234123456789012)
Record count: 1
```
Hooray! it works by doing a hex literal. For sanity sake, it also works with nullable guids:
```
      SELECT x.Id, x.NotNullGuid, x.NullGuid
      FROM GuidsTest AS x
      WHERE x.NullGuid IN (0x47382910291029101029102938475610)
Record count: 1
```

This fix also introduces a common binary formatter for the guids and the binary data type.

Select code:
``` c#
var first = await context.BinaryTest
                .Where(x => shouldExistListOfBinaryArrays.Contains(x.BinaryColumn)).ToListAsync();
```
Schema:
```
CREATE TABLE `BinaryTest` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `binarycolumn` varbinary(10) NOT NULL,
  PRIMARY KEY (`id`)
```

Here is it working when empty:
```
      SELECT x.Id, x.BinaryColumn
      FROM BinaryTest AS x
      WHERE x.BinaryColumn IN (X'')
Binary record count: 1
```

Note that '0x' would not be valid, so I used the X'' syntax. Binary generation with stuff in it:
```
      SELECT x.Id, x.BinaryColumn
      FROM BinaryTest AS x
      WHERE x.BinaryColumn IN (0xCAFEBABE)
Binary record count: 1
```

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/MySqlOldGuidTypeMapping.cs
+++ b/src/EFCore.MySql/Storage/Internal/MySqlOldGuidTypeMapping.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System;
+using JetBrains.Annotations;
+using Pomelo.EntityFrameworkCore.MySql.Utilities;
+
+namespace Microsoft.EntityFrameworkCore.Storage
+{
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    internal sealed class MySqlOldGuidTypeMapping : GuidTypeMapping
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="MySqlOldGuidTypeMapping" /> class.
+        /// </summary>
+        public MySqlOldGuidTypeMapping() : base("binary(16)", System.Data.DbType.Guid) { }
+
+        /// <summary>
+        ///     Generates the SQL representation of a literal value.
+        /// </summary>
+        /// <param name="value">The literal value.</param>
+        /// <returns>
+        ///     The generated string.
+        /// </returns>
+        protected sealed override string GenerateNonNullSqlLiteral([CanBeNull] object value) => ByteArrayFormatter.ToHex(((Guid)value).ToByteArray());
+    }
+}
--- a/src/EFCore.MySql/Storage/Internal/MySqlOldGuidTypeMapping.cs
+++ b/src/EFCore.MySql/Storage/Internal/MySqlOldGuidTypeMapping.cs
@@ -3,6 +3,7 @@
 
 using System;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using Pomelo.EntityFrameworkCore.MySql.Utilities;
 
 namespace Microsoft.EntityFrameworkCore.Storage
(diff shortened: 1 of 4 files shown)


src/EFCore.MySql/Storage/Internal/MySqlOldGuidTypeMapping.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    ///     This API supports the Entity Framework Core infrastructure and is not intended to be used
+    ///     directly from your code. This API may change or be removed in future releases.
+    /// </summary>
+    internal sealed class MySqlOldGuidTypeMapping : GuidTypeMapping
+    {
+        /// <summary>
+        ///     Initializes a new instance of the <see cref="MySqlOldGuidTypeMapping" /> class.
+        /// </summary>
+        public MySqlOldGuidTypeMapping() : base("binary(16)", System.Data.DbType.Guid) { }
```
AndriySvyryd: You should call the constructor that takes a `RelationalTypeMappingParameters` and override the `Clone` methods. This will make this mapping compatible with new EF features.

robbieknuth: Ok updated. I have followed the pattern present in the other classes.

caleblloyd:
@robbieknuth can you add tests for this functionality?

caleblloyd:
OK great.  @AndriySvyryd has the best insight into that.  I'd like to get this into 2.1.0, we will likely be looking to do another RC in 2 weeks, then a final release after that if everything checks out.

caleblloyd:
It still does not have supporting tests, which is why it hasn't been merged yet. @robbieknuth do you plan to add tests still?

adomass:
We really need this fix merged asap since we have to do ugly workarounds now in order to deploy with this fix. @caleblloyd Could you somehow merge it?

caleblloyd:
I went ahead and merged without the tests, could you please validate the behavior is correct in the nightly build:

https://www.myget.org/Package/Details/pomelo?packageType=nuget&packageId=Pomelo.EntityFrameworkCore.MySql

adomass:
Yes it works with nightly build. When do you plan to release new version?

PR #597, opened 2018-06-07
A few more query fixes:

maumar:
- added translation for DateTime.Today,
- fixed range verification for one test,
- added skip justification for 2 others.


maumar:
@AndriySvyryd

Issue #602, opened 2018-06-11
deserialization jsonObject type longer time


huxiany:
Querying the database, processing JsonObject type of deserialization time is too long, resulting in poor performance. 
https://github.com/huxiany/EFCoreIssue

![efcore](https://user-images.githubusercontent.com/28701403/41225619-c84ed0c4-6da1-11e8-9c28-794db4800005.png)

mguinness:
This appears to be a usage issue, in your normal query (getTiles) you aren't filtering (you need to add a where clause).

LiangZugeng:
Thanks for the reply, actually the filter isn't the issue here, the initialization code created 7000 lines of records in the DB and all of them have the same SalesOrderID (578) so it doesn't matter if a where clause is used or not.

The performance bottleneck is at the deserialization of string content of JsonObject, we tested the performance of the two SQL queries (with MaterialSpec or MaterialSpec as null), MySQL returned the dataset to EF Core with almost identical time, and with MaterialSpec it took 50x to 100x time to process the result.

Please kindly reopen the issue and investigate it further, we tried to look into the Polemo.EFCore.MySql source code but we are new to EF Core internal design so we had no much progress pinpointing the root cause.

mguinness:
I suggest you file an issue in the underlying library at [Pomelo.JsonObject](https://github.com/PomeloFoundation/Pomelo.JsonObject).

caleblloyd:
@LiangZugeng if you need performance-critical code you should probably just retrieve as a `string` and write a method to deserialize the field on-demand

Pomelo will Convert any `JsonObject<T>` using reflection when it is read from the database.

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/4c099402e3c044fe0cc0d876d8e73630f2eb36c6/src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs#L96-L111

LiangZugeng:
@caleblloyd thanks for the information.

We spent some time doing the perf profiling and found that the function you mentioned is exactly where the performance issue is, every time a JsonObject needs to be parsed from a string, ConvertWithReflection<T> is called, and every time it constructs the type conversion lambda expression from scratch (including the generation and compilation), this caused a very bad performance.

[...]

PR #606, opened 2018-06-12
Added cache for JsonObject type conversion to improve the performance

LiangZugeng:
To solve the performance issue mentioned in #602 related to JsonObject deserialization, a cache is added to save the compiled lambda expression, this dramatically improves the performance (for 7000 records with Json content the time spent on deserializations dropped from 4.08 seconds to 0.25 seconds, 16x faster).

Also, the previous version of code did JsonObject conversion in a catch block which also had bad performance, the fix uses a explicit strategy for JsonObject determination and conversion to avoid the performance penalty of try-catch block.

All unit and functional test cases were green on a local run.

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs
+++ b/src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs
@@ -1,6 +1,6 @@
-﻿using System;
+using System;
 using System.Collections;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Data.Common;
 using System.IO;
 using System.Linq.Expressions;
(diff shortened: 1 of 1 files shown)


src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs
```diff
@@ -18,6 +19,8 @@ public class WrappedMySqlDataReader : DbDataReader
 
         private DbDataReader _reader;
         private bool _disposed;
+        private Dictionary<string, Delegate> compiledExpressionCache = new Dictionary<string, Delegate>();
```
caleblloyd: Probably makes sense to use a `static` [ConcurrentDictionary](https://msdn.microsoft.com/en-us/library/dd287191(v=vs.110).aspx) here and get rid of the lock

LiangZugeng:
I will submit a new PR.

PR #607, opened 2018-06-13
Added cache for JsonObject type conversion to improve the performance

LiangZugeng:
To solve the performance issue mentioned in #602 related to JsonObject deserialization, a cache is added to save the compiled lambda expression, this dramatically improves the performance (for 7000 records with Json content the time spent on deserializations dropped from 4.08 seconds to 0.25 seconds, 16x faster).
Also, the previous version of code did JsonObject conversion in a catch block which also had bad performance, the fix uses a explicit strategy for JsonObject determination and conversion to avoid the performance penalty of try-catch block.

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs
+++ b/src/EFCore.MySql/Storage/Internal/WrappedMySqlDataReader.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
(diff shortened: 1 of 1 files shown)


caleblloyd:
Fixed in [https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql/2.1.0-rc2-final](2.1.0-rc2-final)

PR #623, opened 2018-06-29
Truncate foreign key and index names to 64 characters in old migrations.

### LINKED ISSUE #590: Old 2.0 Migrations with Foreign Key or Index Identifiers longer than 64 characters does not work properly in 2.1
### Steps to reproduce
Create a database migration that auto generates foreign key constraints or index identifier that is longer than 64 characters.

### The issue
Prior to commit 1191bc26b740f2514ffced8f3eabb46088b68574, foreign key or index identifiers that is greater than 64 characters is truncated when the migration script is being generated by the database provider. Since the ``LimitLength()`` string extension along with it's use has been removed from the codebase, older migrations generated with the 2.0 EF Core provider that happens to have an auto generated FK or index constraint that is longer than 64 characters are now unusable.

As with EF Core 2.1, the ORM now gives [Database Providers](https://github.com/aspnet/EntityFrameworkCore/commit/05d3442f9a6096680d1575beffeea3cf3f764f48) the ability to pre-emptively limit the length of the generated identifiers at the provider level as opposed to being limited when the migration script is generated. Currently, it seems like Pomelo's EF Core provider now takes advantage of this feature without regard to older migrations generated in the 2.0 branch.

### Workaround
Remove all migrations generated with the old 2.0 database provider and start fresh.

AndriySvyryd:
Fixes #590

### DIFF
--- a/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
+++ b/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
@@ -375,6 +375,24 @@ namespace Microsoft.EntityFrameworkCore.Migrations
         }
 
         /// <summary>
+        ///     Builds commands for the given <see cref="CreateIndexOperation" /> by making calls on the given
+        ///     <see cref="MigrationCommandListBuilder" />.
+        /// </summary>
+        /// <param name="operation"> The operation. </param>
+        /// <param name="model"> The target model which may be <c>null</c> if the operations exist without a model. </param>
+        /// <param name="builder"> The command builder to use to build the commands. </param>
+        /// <param name="terminate"> Indicates whether or not to terminate the command after generating SQL for the operation. </param>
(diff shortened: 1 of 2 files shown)


AndriySvyryd:
@ajcvickers FYI

caleblloyd:
@AndriySvyryd can you educate me on how exactly this works?

How does it detect that it is running against an old migration to use truncate?  And what is the new behavior when over 64 characters?

AndriySvyryd:
It doesn't have different behavior for old vs new migrations, the difference is that new migrations will never be generated with identifiers that are more than 64 characters long, EF truncates and uniquifies them beforehand.

caleblloyd:
The "uniquifying" part is important, just truncating can lead to duplicates. So this PR still keeps the "uniquify" behavior for new migrations?

AndriySvyryd:
@caleblloyd correct

PR #670, opened 2018-08-16
escape string literals analogously to parameter values

### LINKED ISSUE #667: String values are not escaped in CONTAINS function
### Steps to reproduce
Executing
```
var groupNames = new[] {"mydomain\\group1", "mydomain\\group2"};
var existingGroups = await dbContext.Groups.AsNoTracking()
    .Where(x => groupNames.Contains(x.Name)).Select(x => x.Name)
    .ToArrayAsync();
```
(with `Name` being a string field) resolves to 
```
SELECT `x`.`Name`
FROM `Groups` AS `x`
WHERE `x`.`Name` IN ('mydomain\group1', 'mydomain\group2')
```
when the expected resulting SQL query would be 
```
[...] WHERE `x`.`Name` IN ('mydomain\\group1', 'mydomain\\group2')
```

### The issue
In my case, this only results in zero rows regardless whether those values exist in the database. 
Potentially this opens vulnerabilites for SQL injection attacks whenever developers rely on input being escaped by the EF connector (i. e. a lot, I suppose). I have not checked whether other queries are affected as well or if this behaviour is limited to contains function calls.

### Further technical details
Pomelo.EntityFrameworkCore.MySql version: 2.1.0

wiredbarb:
Fixes #667 

Issue: Constant string values are not escaped (except for occurances of `'` being replaced with `''`). 
Since variable values contained in [`InExpression`](https://github.com/aspnet/EntityFrameworkCore/blob/release/2.2/src/EFCore.Relational/Query/Expressions/InExpression.cs)s will be resolved to constants when those are expanded during query generation, this also affects use cases dealing with input other than hard coded strings.

Solution: Handle constant string values the same way as parameter string values (see escaping in [`MySqlParameter`](https://github.com/mysql-net/MySqlConnector/blob/62690094a865b68ba15136c1e5d9304df878757d/src/MySqlConnector/MySql.Data.MySqlClient/MySqlParameter.cs#L190)).

Parameter values in insert/update operations will still not be escaped correctly with NO_BACKSLASH_ESCAPES, but I think this really is an edge case and it will be obsolete as soon as prepared statements are used - which seems to be in the making already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,19 @@ namespace YourNamespace // replace "YourNamespace" with the namespace of your ap
     }
 }
 ```
+#### NO_BACKSLASH_ESCAPES
+The escaping style can be configured to support SQL mode [NO_BACKSLASH_ESCAPES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_backslash_escapes) by adding `.DisableBackslashEscaping()` to your MySQL option configuration.
+```csharp
+mysqlOptions =>
+{
+    mysqlOptions
+    .ServerVersion(new Version(5, 7, 17), ServerType.MySql); // replace with your Server Version and Type
+    //further MySQL option configurations go here
+    .DisableBackslashEscaping();
+}
+```
+**PLEASE NOTE** This option does not set the SQL mode and will not check if it matches the actual database configuration either. Do not use this option unless you are absolutely sure it will always be set when queries are performed on the context. 
+As of today, it also **does not apply** to parameter values in insert or update statements; do not use it when the context is used to update entries.
 
 View our [MySql Provider Configuration Options Wiki Page](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/wiki/MySql-Provider-Configuration-Options) for a complete list of supported options.
 
--- a/src/EFCore.MySql/Infrastructure/Internal/IMySqlOptions.cs
+++ b/src/EFCore.MySql/Infrastructure/Internal/IMySqlOptions.cs
@@ -13,5 +13,6 @@ namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal
         CharSetBehavior CharSetBehavior { get; }
         CharSetInfo AnsiCharSetInfo { get; }
         CharSetInfo UnicodeCharSetInfo { get; }
+        bool NoBackslashEscapes { get; }
     }
 }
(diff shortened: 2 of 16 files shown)


mguinness:
@wiredbarb If SQL mode [NO_BACKSLASH_ESCAPES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_backslash_escapes) is used this breaks functionality.

wiredbarb:
Currently, the behaviour is being inconsistent. 
```
var name = "mydomain\group'xy";
_context.DbSet<Group>().Where(x => x.Name == name).Count()
```
already (correctly) results in
```
SELECT COUNT(*)
FROM `Groups` AS `x`
WHERE (`x`.`Name` = 'mydomain\\group\'as') 
```
This does not work with NO_BACKSLASH_ESCAPES on either.

It ***will*** break functionality in scenarios using constant strings that are already escaped, though.

In order to consistently respect NO_BACKSLASH_ESCAPE mode settings, an additional query would have to be made before every request containing string literals, checking whether backslashes should be escaped or not. (Even if this is set session-wise you cannot guarantee it is not changed in the meantime - in a stored procedure for instance.) And since parameters are generally re-weaved into the query by MySqlConnector, this would have to be applied to both parameter and constant values.

wiredbarb:
Maybe an additional option parameter allowing to opt out of backslash escapes in general would be an alternative for users who can warrant that NO_BACKSLASH_ESCAPE is active?

mguinness:
There used to be code to determine server capabilities so maybe `SELECT @@GLOBAL.sql_mode` could be added to that, otherwise it could be added to `MySqlOptionsExtension` class?  @caleblloyd Any suggestions?

wiredbarb:
Now, [this commit ](https://github.com/ahdde/Pomelo.EntityFrameworkCore.MySql/commit/3a0e416f6472b5852bf6ab6eeffb671f90842a97) introduces a general opt-out configuration. 

The cleaner solution would be to _actually_ send parameters (which would have to be fixed upstream in MySqlConnector anyway), but this would not solve the IN-issue, because EF Core still treats all list items as constants. Plus, I have not yet investigated whether this is supported in MySQL payload at all.

I have not added any unit / integration tests yet, which I will rectify when/if you approve of this implementation.
Also, an initial check whether NO_BACKSLASH_ESCAPES is actually set on the database when the opt-out-configuration is used might be worth consideration, but I am not sure where to put this.

mguinness:
It'll be best to wait for caleblloyd's input as he'll be best placed to answer your questions and he's also a contributor on MySqlConnector repo.  Thanks for your help.

mguinness:
It's a pretty basic change and your implementation seems sound (so skip NO_BACKSLASH_ESCAPES check).  Please go ahead and add unit / integration tests.  Can also add document the NoBackslashEscapes option in README?

collinbarrett:
Looking forward to this being released. If I understand correctly, until it is, it is generally somewhat unsafe to use EF `.Contains()` if the data is expected to have a variety of special characters?

wiredbarb:
@collinbarrett Yes. Until it is, replacing `\` with `\\` in string parameter values to be used with EF `Contains() ` (i. e. the `lines` list ) should be a sufficient workaround.

PR #779, opened 2019-04-22
Connection resilience when unable to connect or timeout

BartjeD:
- Adjsuted ShouldRetryOn() to also trigger on conditions `UnableToConnectToHost` and `CommandTimeoutExpired`. And switched to using error code enum instead of integers.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -6,7 +6,7 @@
   <PropertyGroup Label="Package Versions">
     <!-- EFCore.MySql Dependencies -->
     <MicrosoftEntityFrameworkCoreRelationalVersion>2.2.0</MicrosoftEntityFrameworkCoreRelationalVersion>
-    <MySqlConnectorVersion>0.49.2</MySqlConnectorVersion>
+    <MySqlConnectorVersion>0.51.1</MySqlConnectorVersion>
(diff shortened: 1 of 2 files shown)


mguinness:
Won't you need to bump up version to 0.51.0 for MySqlConnector?

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/c58e00c7b3b6bc8f2eeffebd4d10f2e1ff16b790/build/dependencies.props#L9

BartjeD:
For the UnableToConnectToHost it shouldn't be necessary, that exception was already returned and wasn't handled resliently. Which is what the original request was about.

But to be honest I had forgotten about the package update after looking at the exception codes in the MySqlConnector repo. :x: 

Looking at the MySqlConnector packages shows that there's v. 0.51.1, so I'll update to that. Because it doens't appear to be a breaking update above 0.51.0

BartjeD:
@mguinness 

The version has been bumped to 0.51.1.

The MySqlConnector changelog doesn't appear to list any breaking changes going to 0.50 and 0.51.
But I can't judge whether there's anything there, apart from looking at the unittests. Could you review the connector changelog? :)

https://github.com/mysql-net/MySqlConnector/releases/tag/0.50.0
https://github.com/mysql-net/MySqlConnector/releases/tag/0.51.0

Issue #745, opened 2019-02-19
Trying to setup build env


jrocha:
### Steps to reproduce
Just cloned this repo to a Win10 and tried to setup a build environment.

### The issue
Looks it is not happy run on MariaDB 10.3

```
Exception message: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '>> "$[0]") VIRTUAL NULL,
    `Email` VARCHAR(63) GENERATED ALWAYS AS (`ContactI' at line 3
Stack trace:
```

### Further technical details
C:\data\Pomelo.EntityFrameworkCore.MySql\test\EFCore.MySql.IntegrationTests\scripts>rebuild.cmd
Dropping database 'pomelo_test'.
Successfully dropped database 'pomelo_test'.
Done. To undo this action, use 'ef migrations remove'
C:\data\Pomelo.EntityFrameworkCore.MySql\test\EFCore.MySql.IntegrationTests\Migrations\20190219180244_initial.cs
C:\data\Pomelo.EntityFrameworkCore.MySql\test\EFCore.MySql.IntegrationTests\Migrations\20190219180244_initial.Designer.cs
C:\data\Pomelo.EntityFrameworkCore.MySql\test\EFCore.MySql.IntegrationTests\Migrations\AppDbModelSnapshot.cs
MySql.Data.MySqlClient.MySqlException (0x80004005): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '>> "$[0]") VIRTUAL NULL,
    `Email` VARCHAR(63) GENERATED ALWAYS AS (`ContactI' at line 3 ---> MySql.Data.MySqlClient.MySqlException (0x80004005): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '>> "$[0]") VIRTUAL NULL,
    `Email` VARCHAR(63) GENERATED ALWAYS AS (`ContactI' at line 3
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs:line 43
[...]

PR #780, opened 2019-04-27
Bart/unquoting mariadb fix

BartjeD:
- GeneratedContactMeta: Besides checking for AppVeyor there is now also a check for MariaDb. If either, return.

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/745

### DIFF
--- a/test/EFCore.MySql.IntegrationTests/Tests/Attributes.cs
+++ b/test/EFCore.MySql.IntegrationTests/Tests/Attributes.cs
@@ -1,11 +1,12 @@
-﻿using Xunit;
+﻿using Pomelo.EntityFrameworkCore.MySql.Storage.Internal;
+using Xunit;
 
 namespace Pomelo.EntityFrameworkCore.MySql.IntegrationTests.Tests
 {
-	public class SkipAppVeyorFact : FactAttribute {
+	public class SkipAppVeyorAndMariaDbFact : FactAttribute {
 
-		public SkipAppVeyorFact() {
-			if(AppConfig.AppVeyor) {
+		public SkipAppVeyorAndMariaDbFact() {
+			if(AppConfig.AppVeyor  || new ServerVersion(AppConfig.Config["Data:ServerVersion"]).Type == Infrastructure.ServerType.MariaDb) {
 				Skip = "Test does not work with AppVeyor's MySQL version";
 			}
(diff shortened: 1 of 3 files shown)


test/EFCore.MySql.IntegrationTests/Tests/Attributes.cs
```diff
@@ -1,11 +1,12 @@
-﻿using Xunit;
+﻿using Pomelo.EntityFrameworkCore.MySql.Storage.Internal;
+using Xunit;
 
 namespace Pomelo.EntityFrameworkCore.MySql.IntegrationTests.Tests
 {
-	public class SkipAppVeyorFact : FactAttribute {
+	public class SkipAppVeyorAndMariaDbFact : FactAttribute {
```
caleblloyd: Instead of tacking on `SkipAppVeyorAndMariaDbFact`, leave `SkipAppVeyorFact` and add a `SkipMariaDbFact`

test/EFCore.MySql.IntegrationTests/Tests/Models/GeneratedTypesTest.cs
```diff
@@ -11,7 +11,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.IntegrationTests.Tests.Models
 {
 	public class GeneratedTypesTest
 	{
-		[SkipAppVeyorFact]
+		[SkipAppVeyorAndMariaDbFact]
```
caleblloyd:
```
[SkipAppVeyorFact]
[SkipMariaDbFact]
```

BartjeD: @caleblloyd xUnit2 doesn't allow multiple test attributes on a method. I wanted them seperate too :/
`Message: System.InvalidOperationException : Test method 'Pomelo.EntityFrameworkCore.MySql.IntegrationTests.Tests.Models.GeneratedTypesTest.TestGeneratedContact' has multiple [Fact]-derived attributes
`

caleblloyd: Ah I did not know.  Makes sense how you have it then

Issue #799, opened 2019-07-16
DateTimeOffset cannot be used as an Identity Column


crozone:
### Steps to reproduce
Create an entity with a `DateTimeOffset` property, and mark it as an identity column. For example:

```
[DatabaseGenerated(DatabaseGeneratedOption.Identity)]
        public DateTimeOffset CreatedTime { get; set; }
```

Generate a migration. Notice the entry for the column has no annotation:

```
migrationBuilder.AddColumn<DateTimeOffset>(
                name: "CreatedTime",
                table: "SomeTable",
                nullable: false,
                defaultValue: new DateTimeOffset(new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified), new TimeSpan(0, 0, 0, 0, 0)));
```

However, when this is repeated with a `DateTime` field, the annotation is created:

```
[DatabaseGenerated(DatabaseGeneratedOption.Identity)]
        public DateTime CreatedTime { get; set; }
```

...

```
migrationBuilder.AddColumn<DateTime>(
                name: "CreatedTime",
                table: "SomeTable",
                nullable: false,
                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified))
                .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn);
```

### The issue
The ClrTypes `DateTime` and `DateTimeOffset` both map to `datetime(6)`, so it is expected that they both have the same ability to be used as an Identity Column. Both types already support `MySqlValueGenerationStrategy.ComputedColumn`, for example when using the `[Timestamp]` attribute.

The issue appears to be in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/2c5a931ea2b5170e29ac7f1b8f36cc7cd0de2728/src/EFCore.MySql/Metadata/MySqlPropertyAnnotations.cs#L343-L355

`IsCompatibleIdentityColumn` only checks for `DateTime`, while `IsCompatibleComputedColumn` checks for both `DateTime` and `DateTimeOffset`.

This was updated in commit https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/commit/e988d1afaf6e29215456a2b33b9a3a23ce96c385, it appears intentional, but why?

[...]

PR #800, opened 2019-07-16
Fix datetimeoffset identity column

crozone:
Trivial fix for https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/799

This allows the `DateTimeOffset` columns to function with `ValueGeneratedOnAdd()`, `ValueGeneratedOnAddOrUpdate()`, and `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]`, in exactly the same way that `DateTime` does.


caleblloyd:
Thanks for your contribution

Issue #801, opened 2019-07-16
DateTime default value is incorrect when using DatabaseGenerated Identity or Computed columns


crozone:
### Steps to reproduce
Create an entity with `DateTime` properties, and decorate them with either `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]` or `[DatabaseGenerated(DatabaseGeneratedOption.Computed)]`:

```
[DatabaseGenerated(DatabaseGeneratedOption.Identity)]
public DateTime CreatedTime { get; set; }

[DatabaseGenerated(DatabaseGeneratedOption.Computed)]
public DateTime UpdatedTime { get; set; }
```
Generate a migration for the change.

Notice that the migration has produced `AddColumn` methods with a default value of `defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified))`:

```
migrationBuilder.AddColumn<DateTime>(
                name: "CreatedTime",
                table: "SomeTable",
                nullable: false,
                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified))
                .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn);
```
...
```
migrationBuilder.AddColumn<DateTime>(
                name: "UpdatedTime",
                table: "SomeTable",
                nullable: false,
                defaultValue: new DateTime(1, 1, 1, 0, 0, 0, 0, DateTimeKind.Unspecified))
                .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.ComputedColumn);
```

Then, script the migration. Notice that the generated SQL uses the default value listed in the migrations file, instead of the desired `CURRENT_TIMESTAMP(6)` SQL.

What we want from `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]`:

``ALTER TABLE `SomeTable` ADD `CreatedTime` datetime(6) NOT NULL DEFAULT CURRENT_TIMESTAMP(6);``

What we get from `[DatabaseGenerated(DatabaseGeneratedOption.Identity)]`:

``ALTER TABLE `SomeTable` ADD `CreatedTime` datetime(6) NOT NULL;``

What we want from `[DatabaseGenerated(DatabaseGeneratedOption.Computed)]`:

[...]

PR #802, opened 2019-07-16
Potential fix for issue-801

crozone:
https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/801

Use IdentityColumn value generation if default value is the clrType default, instead of just null (which only works for reference types)

Effectively this allows the `.Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn)` annotation to override the `defaultValue` parameter if that default value is equal to the default value for the CLR type. This means that any column which maps to a struct CLR type can use `MySqlValueGenerationStrategy.IdentityColumn`.

However, this might be a bit of a kludge. Ideally, the migration `migrationBuilder.AddColumn()` generated would have `defaultValue` set to `null` in this case, avoiding the need for this kludge. However I'm unsure of the changes required to make that happen.


crozone:
Closing, this is probably not the way to go about this.

PR #806, opened 2019-07-30
add a generic Like method to support more type in `LIKE` expression

ElderJames:


src/EFCore.MySql/MySqlDbFunctionsExtensions.cs
```diff
@@ -450,5 +450,54 @@ public static class MySqlDbFunctionsExtensions
...
+        ///     </para>
+        ///     <para>
+        ///         Note that if this function is translated into SQL, then the semantics of the comparison will
+        ///         depend on the database configuration. In particular, it may be either case-sensitive or
+        ///         case-insensitive. If this function is evaluated on the client, then it will always use
+        ///         a case-insensitive comparison.
+        ///     </para>
+        /// </summary>
+        /// <param name="_">The DbFunctions instance.</param>
+        /// <param name="matchExpression">The string that is to be matched.</param>
```
bgrainger: This comment seems wrong; `matchExpression` is not a `string`.

ElderJames: Sorry,I've fixed it.

src/EFCore.MySql/MySqlDbFunctionsExtensions.cs
```diff
@@ -450,5 +450,54 @@ public static class MySqlDbFunctionsExtensions
...
+        ///         a case-insensitive comparison.
+        ///     </para>
+        /// </summary>
+        /// <param name="_">The DbFunctions instance.</param>
+        /// <param name="matchExpression">The property of entity that is to be matched.</param>
+        /// <param name="pattern">The pattern which may involve wildcards %,_,[,],^.</param>
+        /// <returns>true if there is a match.</returns>
+        public static bool Like<T>(
+            [CanBeNull] this DbFunctions _,
+            T matchExpression,
```
bgrainger: It appears (from https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/806/files#diff-8764c7d56fa4b46141c410f9975704ebR63) that this can be specialised on `double`, `decimal`, etc. `ToString` is not safe to call in non-U.S.-English cultures, because it can format `1.23` as `"1,23"`, which won't be accepted by MySQL Server. 

Can `T` here be a type that needs culture-invariant formatting? If so, you might need to use `IConvertible`.

src/EFCore.MySql/MySqlDbFunctionsExtensions.cs
```diff
@@ -450,5 +450,54 @@ public static class MySqlDbFunctionsExtensions
...
+        ///         case-insensitive. If this function is evaluated on the client, then it will always use
+        ///         a case-insensitive comparison.
+        ///     </para>
+        /// </summary>
+        /// <param name="_">The DbFunctions instance.</param>
+        /// <param name="matchExpression">The property of entity that is to be matched.</param>
+        /// <param name="pattern">The pattern which may involve wildcards %,_,[,],^.</param>
+        /// <returns>true if there is a match.</returns>
+        public static bool Like<T>(
+            [CanBeNull] this DbFunctions _,
```
bgrainger: Nitpick: It feels very strange to call this parameter `_` (which implies a "discard"), but then use it immediately on line 474. `functions` might be a better name.

I do see that this has precedence in this file, but I don't think it's a good precedent to follow.

src/EFCore.MySql/MySqlDbFunctionsExtensions.cs
ElderJames: I have already used `IConvertible` in the latest commit.

src/EFCore.MySql/MySqlDbFunctionsExtensions.cs
ElderJames: Now I no longer use `_` immediately 😄

ElderJames:
How is it going on?

lauxjpn:
@bgrainger Thanks for reviewing this.
@ElderJames Thank you for your contribution and sorry for the delay! This will be part of the 2.2.6 release and will also be cherry-picked into the 3.0.0 branch and updated accordingly.

ElderJames:
Thanks！

Issue #776, opened 2019-04-16
Stored Procedure and Migrations


agustinbcu01:
### Steps to reproduce
I create an empty migration on the UP Method, I write this.
```
  protected override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.Sql(@"
                   
                    DROP PROCEDURE IF EXISTS GetCurrentGeneration;
                       CREATE  PROCEDURE GetCurrentGeneration(IN timeOut INT, IN createBy char(255))
                        BEGIN
                                 DECLARE currentTimeStamp TIMESTAMP;
                                 DECLARE generationsCount INT;
                                 DECLARE sqlUpdated INT;
             
                                 SET currentTimeStamp = UTC_TIMESTAMP();

                        SET sqlUpdated = 1; -- SQL_SAFE_UPDATES;     
                        SET SQL_SAFE_UPDATES = 0;
                        UPDATE generetions 
                        SET 
                            status = 3
                        WHERE
                            DATE_ADD(UpdatedAt,
                                INTERVAL timeOut SECOND) < currentTimeStamp
                                AND Status IN (1 , 2);	
            
                                SET   SQL_SAFE_UPDATES = sqlUpdated;
                   

    			                    INSERT INTO generetions
                                    ( StartAt, Status, ImagesCount, VideosCount, CreatedBy, CreatedAt, UpdatedBy, UpdatedAt)
                                    SELECT * FROM (SELECT currentTimeStamp, 1 as a, 0 as b, 0 as c, createBy as d, currentTimeStamp as e, createBy as f,currentTimeStamp as g) AS tmp
                                    WHERE NOT EXISTS (SELECT id from generetions WHERE   Status IN (1 , 2)) limit 1;
         
              
                        SELECT 
                            *
                        FROM
                            generetions
                        WHERE
                            Status IN (1 , 2)
                        ORDER BY UpdatedAt DESC;
        
                   
         
                        END
            ");
        }
```
and I generate the migration sql script.
```

DROP PROCEDURE IF EXISTS MigrationsScript;
DELIMITER //
CREATE PROCEDURE MigrationsScript()
BEGIN
    IF NOT EXISTS(SELECT 1 FROM `__EFMigrationsHistory` WHERE `MigrationId` = '20190411134055_AddGenerations_sp') THEN
    
                       
                       DROP PROCEDURE IF EXISTS GetCurrentGeneration;
[...]

PR #810, opened 2019-08-03
Extract primary key related SQL functions into migration header

igrmk:
This one fixes idempotent migration scripts if they contains table renames or primary key droppings or creations.
The bug with idempotent migrations is described here https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/776
This fixes nonidempotent migration scripts with such the actions as well. Currently they just don't work because of delimiters.

Rather than creating and dropping related procedures every time whenever we change primary key this change extracts them into the beginning of migration script.


lauxjpn:
Merging this PR should be hold for the moment to check whether migrations can be implemented without the use of stored procedures: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/776#issuecomment-537545623

If so, then this fix should be implemented without SPs as well.

lauxjpn:
As it turns out, *idempotent* migrations cannot be implemented without the use of stored procedures at this point in time.

src/EFCore.MySql/Migrations/Internal/MySqlMigrator.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+        }
+
+        public override string GenerateScript(
+            string fromMigration = null,
+            string toMigration = null,
+            bool idempotent = false)
+        {
+            return Header() + base.GenerateScript(fromMigration, toMigration, idempotent) + Footer();
+        }
```
lauxjpn: I think the appropriate stored procedures should only be added if they are actually needed by the generated script. Otherwise this could end up as a long list of mostly unused stored procedures in the future.

lauxjpn:
Preferably merge this into `3.0.0-wip` (otherwise use `2.2.6-wip`).

src/EFCore.MySql/Migrations/Internal/MySqlMigrator.cs
igrmk: Done

lauxjpn:
Don't worry about the failing CI. This has probably nothing to do with your PR and will be fixed shortly.

lauxjpn:
So the CI has been fixed 6 days ago and I have retriggered the CI build for this PR.
@igrmk There are build errors, which need to be fixed:

```
Build FAILED.

Migrations\Internal\MySqlMigrator.cs(29,15): error CS7036: There is no argument given that corresponds to the required formal parameter 'commandLogger' of 'Migrator.Migrator(IMigrationsAssembly, IHistoryRepository, IDatabaseCreator, IMigrationsSqlGenerator, IRawSqlCommandBuilder, IMigrationCommandExecutor, IRelationalConnection, ISqlGenerationHelper, ICurrentDbContext, IDiagnosticsLogger<DbLoggerCategory.Migrations>, IDiagnosticsLogger<DbLoggerCategory.Database.Command>, IDatabaseProvider)' [D:\a\1\s\src\EFCore.MySql\EFCore.MySql.csproj]
Migrations\Internal\MySqlMigrator.cs(14,34): error EF1001: Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator is an internal API that supports the Entity Framework Core infrastructure and not subject to the same compatibility standards as public APIs. It may be changed or removed without notice in any release. [D:\a\1\s\src\EFCore.MySql\EFCore.MySql.csproj]
Migrations\Internal\MySqlMigrator.cs(84,28): error EF1001: Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator is an internal API that supports the Entity Framework Core infrastructure and not subject to the same compatibility standards as public APIs. It may be changed or removed without notice in any release. [D:\a\1\s\src\EFCore.MySql\EFCore.MySql.csproj]
    0 Warning(s)
    3 Error(s)
```

igrmk:
@lauxjpn Build errors should be fixed now. Internal API was changed little bit in 3.0.0

lauxjpn:
@igrmk Something seems broken here with this PR and the CI. Please try to open another PR and see if if works there.

igrmk:
I created this one https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/883

lauxjpn:
PR #883 has been merged.

PR #820, opened 2019-09-11
Upgrade to 2.2.6 and fix tests

### LINKED ISSUE #571: Query: invalid query generated for some cases requiring cast to certain types (e.g. double)
Casting is only supported for handful of types. Currently MySql tries to convert to other types, e.g double to decimal. The problem with this approach is that we get cast exception during materialization (expecting double, got decimal). This could potentially be improved by using value converters. Another approach would be to client eval, but that causes a lot of client eval (perf hit) and exposes a lot of bugs in query pipeline.

lauxjpn:
This PR upgrades the provider (and especially the tests) to 2.2.6.

I also took the time to fix some existing issues, like the CAST() problems (fixes #571 ). There should be added additional tests in the future, to ensure that all possible CAST() scenarios work now. If anything needs to be added, it can be implemented either using SQL/Expressions in `MySqlQuerySqlGenerator.VisitExplicitCast()` or as a `DbDataReader` injection in `MySqlConverterDataReader `.

I ran the test before and after, with the following results:

EF (tests) | Total | Passed | Failed | Skipped
-----------|-------|--------|--------|--------
2.2.0 (2.1.4) | 7483 | 7261 | 30 | 192
2.2.6 (2.2.6) | 8886 | 8743 | 12 | 131

None of the failing tests of this PR were working before, so I think we are good for now. Some of the remaining failing tests can be implemented, I just didn't get to it yet.

I would like to base the upcoming 3.0 upgrade on top of this PR, so a review sooner than later is appreciated.

For context, check the individual commits.

@mguinness 
@caleblloyd


AndriySvyryd:
While it's ok for 2.2.6 generally the DataReader shouldn't be wrapped as it can cause maintainability issues. For 3.x this issue should be fixed by overriding parts of the default query pipelines.

lauxjpn:
@AndriySvyryd 
It is definitely not an ideal solution. I just couldn't figure out a way to make this work with the current query pipeline on short notice.

For example, for a user that converts an `Int32` to a `Single` in his LINQ-query, we basically need to convert the value to a `decimal` in SQL, due to limited CAST() support in MySQL.
As far as I can see, to not override the `DbDataReader`, we would need to change the expression that EFCore uses to read the expected `Single` value from the `DbDataReader` in such a way, that it would call `DbDataReader.GetDecimal()` instead of `DbDataReader.GetFloat()`, and then implement the conversion from `Decimal` back to `Single` after the values has been read.

Any pointers in the right direction (or to the right class) for a proper 3.0 implementation would be helpful.

AndriySvyryd:
@lauxjpn As I said, for 2.2.6 it's a good solution. In 3.0 the query pipeline is completely different and has more granular type mapping. It would be best to discuss the 3.0 approach separately from this PR with me, @smitpatel, @roji and @bgrainger (and anyone else who's interested)

bgrainger:
The upcoming MySqlConnector 0.57.0 will support automatically converting a `DECIMAL` column to `float` if `MySqlDataReader.GetFloat()` is called (and `GetDouble`, too): https://github.com/mysql-net/MySqlConnector/issues/664

Would upgrading the connector be sufficient to solve this problem? (There are commits on this branch to prepare for that upgrade: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/commits/2.2.1.)

bgrainger:
Looks like there might be a need to add a conversion from `BIGINT` in `GetByte()`. (`GetByte()` doesn't currently support any narrowing conversions and I'd have to look into why that is. Looks like it might have been for consistency with MySql.Data and SqlClient, but it's different from Npgsql and Sqlite: https://mysql-net.github.io/AdoNetResults/#GetByte_throws_for_one_Int16.)

caleblloyd:
@lauxjpn thanks for the PR.  I do not currently have the time to dedicate to reviews/performing upgrades, I believe others in the community do though!  I created a new base branch `2.2.6-wip` and `3.0.0-wip` and gave you write access to them.  Let me know if any issues merging PRs into them.

I am happy to help with maintenance tasks such as merges to master and publishing releases once they are ready.

lauxjpn:
What is the correct version number for this upgrade? Should it correlate with EFCore and be 2.2.6, or is there a different approach used here?

lauxjpn:
I resolved the discrepancies between `Convert.ToInt32` (and similar calls) and CAST() and updated the version to `2.2.6-wip`.
I am going to merge now.

Issue #821, opened 2019-09-15
Test suite behavior for unsupported features


lauxjpn:
How do we want to handle unsupported features in the test suit?
Do we want to disable the failing test
```C#
[ConditionalTheory(Skip = "EXCEPT is not natively supported by MySQL.")]
public override Task Except(bool isAsync)
{
    return base.Except(isAsync);
}
```
or do we want to override the test to explicitly ensure that the test result is the one we expect (e.g. a thrown exception of some type), which would lead to a green test if it fails as expected?
```C#
public override Task Except(bool isAsync)
{
    // EXCEPT is not natively supported by MySQL.
    return Assert.ThrowsAsync<InvalidOperationException>(() => base.Except(isAsync));
}
```

lauxjpn:
The Sqlite provider does handle those cases as part of the test (instead of skipping the test):
https://github.com/aspnet/EntityFrameworkCore/blob/817aaea6ca1a24a29e8e9f3a7b2dd3748956d942/test/EFCore.Sqlite.FunctionalTests/Query/SimpleQuerySqliteTest.cs#L25-L33

```c#
 // SQLite client-eval
public override async Task Sum_with_division_on_decimal(bool isAsync)
{
    Assert.StartsWith(
        "The LINQ expression",
        (await Assert.ThrowsAsync<InvalidOperationException>(
            () => base.Sum_with_division_on_decimal(isAsync)))
        .Message);
}
```
This seems to way to go then.

PR #818, opened 2019-09-05
[WIP] Feature/netcore3.0

twsI:
This is the work in progress for the support of .NET Core 3.
There is still a lot to do: tests are not working yet, quite a lot of things are unclear to me and it might need a general cleanup. But a lot of API calls are updated already and I hope, that this work can contribute towards the next release somehow. I am unsure whether I can continue to work on this or not and I would love to see someone take over.

Further discussion can be found within this issue #797 

In case you'd like to thank me for my work, please like this issue https://github.com/aspnet/EntityFrameworkCore/issues/11597

PR #822, opened 2019-09-18
Upgrade to 3.0.0

lauxjpn:
This PR updates the provider and most tests to EF Core 3.0.
It is based on #820 and @twsl's extensive work from #818.

It should be noted for major updates in the future, that directly merging the SQL Server files (or parts of it) in this repository, is not the optimal way to proceed here, because it leads to many subtle changes in `SqlFunctionExpression` parameter orders and small behavior differences, that are hard to track down.
Instead, the updates should be triggered by failing tests, checking the [providers log](https://docs.microsoft.com/en-us/ef/core/providers/provider-log), comparing changes between different versions of the Sqlite and NpSql providers and applying the necessary changes manually.

On the plus side, having done this now once, the file and code structure is now much closer to those of other providers and should be easier to maintain in the future. I suggest keeping it this way.

The following table shows the current test metrics:

Version | Total | Passed | Failed | Skipped
-----------|-------|--------|--------|--------
2.2.6 | 8886 | 8743 | 12 | 131
3.0.0 | 14102 | 13389 | 237 | 476

Some tests that passed in the previous releases fail now by design/default, due to missing native support in MySQL for an underlying feature, that would implicitly trigger client eval in previous versions (see #821 for further discussion).
These include all `EXCEPT`, `INTERSECT` and `OUTER APPLY` queries.
Some test queries reference columns from an outer table and fail.
It might be possible to implement the latter and the `OUTER APPLY` functionality (at least for some cases) with the new `LATERAL` keyword (MySQL 8.0.14). This needs further research.

NO_BACKSLASH_ESCAPES is still disabled.

I haven't checked yet, if there are additional test classes to implement for 3.0.

The dependency versions should be updated shortly to public and stable releases.

I kept the original commit structure intact, to make it easier to review. All or most commits should be squashed when merging.

Everybody should feel encoraged to test/review this PR, so we can get this out there for people to test their own projects against and provide additional feedback.

I will continue fixing tests and isolating those, that depend on not implemented features.

I updated the CI files to support 3.0.


caleblloyd:
Thanks for all of your work I'd like to get this merged into `3.0.0-wip`

I changed CI to use Azure DevOps in #823.  Can you merge master into your branch and update CI?  

I was hoping we could move to full test coverage on Azure DevOps, but if you want to make it test a subset as `master` does now we can work that in a later PR

Once this is in the WIP branch I can setup Azure Devops to deploy nightlies off the WIP branch so people can start testing it out

lauxjpn:
I will be merging this in 6 to 10 hours.

lsvMoretti:
Would this fix my issue I have here?

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/825

lauxjpn:
@lsvMoretti Probably. Using a 2.2.x provider with EF Core 3.0 will not work. Your error seems to be related to some dependency mismatches between EF Core 3.0 and the provider.
Tryout the new 3.0 branch once merged and let us know.

lauxjpn:
@caleblloyd Do you want me to squash all commits during the merge, squash them manually into a reasonable amounts of commits (~10) before merging, or leave them as is and merge without squashing at all?

caleblloyd:
Just hit the squash and merge button on the PR

lauxjpn:
I updated the Azure DevOps yaml to use .NET Core 3.0 and skip all tests for now, to merge this quickly as a WIP.
After merging, I will open another issue for keeping track on the newly implemented features, their supporting test, missing features and the remaining failing tests.

At this point in time, the test metrics look like this: (updated shortly)

MySQL | Total | Passed | Failed | Skipped
-----------|-------|--------|--------|--------
5.7.27 |  |  |  | 
8.0.17 |  |  |  | 

The difference comes mainly from `PARTITION BY` and the newly supported `LATERAL` support in MySQL 8 (though some tests fail due to MySQL bugs I reported).

Careful with migrations. They might not work fully yet.

build/dependencies.props
```diff
@@ -5,26 +5,34 @@
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
     <!-- EFCore.MySql Dependencies -->
-    <MicrosoftEntityFrameworkCoreRelationalVersion>2.2.0</MicrosoftEntityFrameworkCoreRelationalVersion>
-    <MySqlConnectorVersion>0.51.1</MySqlConnectorVersion>
+    <MicrosoftEntityFrameworkCoreRelationalVersion>3.0.0-rc1.19452.2</MicrosoftEntityFrameworkCoreRelationalVersion>
+    <MySqlConnectorVersion>0.57.0-rc1</MySqlConnectorVersion>
```
bgrainger: Thanks! I plan to have a stable version out very soon; just waiting for the final SDK.

lauxjpn: Looking forward to it.
I might open an issue about the `NO_BACKSLASH_ESCAPES` support in conjunction with parameters for MySqlConnect tomorrow. Just have to do a bit more research on that before.

PR #823, opened 2019-09-19
Use Azure DevOps

pomelo-efcore-sa:
Changes Linux/Windows builds to AZDO


azure-pipelines.yml
```diff
@@ -0,0 +1,87 @@
...
+        cp test\EFCore.MySql.IntegrationTests\appsettings.ci.json test\EFCore.MySql.IntegrationTests\appsettings.json
+        cp test\EFCore.MySql.IntegrationTests\config.json.example test\EFCore.MySql.IntegrationTests\config.json
+        .\run.ps1 install-tools
+        .\dotnet-run.ps1 .\test\EFCore.MySql.IntegrationTests\scripts\rebuild.ps1
+      displayName: Setup
+    - pwsh: .\build.cmd "/p:SkipTests=true"
+      displayName: Build Project
+    - pwsh: .\dotnet-run.ps1 dotnet test test\EFCore.MySql.Tests
+      displayName: Tests
+    - pwsh: .\dotnet-run.ps1 dotnet test -c Release test\EFCore.MySql.FunctionalTests --filter "FullyQualifiedName~Pomelo.EntityFrameworkCore.MySql.FunctionalTests.MigrationSqlGeneratorMySqlTest|FullyQualifiedName~Pomelo.EntityFrameworkCore.MySql.FunctionalTests.MigrationSqlGeneratorMySql55Test"
```
lauxjpn: What is the reason for only testing `MigrationSqlGeneratorMySqlTest` and `MigrationSqlGeneratorMySql55Test` of the `FunctionalTests` suite?

caleblloyd: Historically TravisCI couldn't handle the full test suite (it would freeze) so I had to just pick some to run as part of the CI suite.

I'd like to move to the full test suite in CI if it's feasible. That probably means ignoring all of the failing tests though

PR #828, opened 2019-09-21
Use Arcade for Build

caleblloyd:
.NET Core packages now use https://github.com/dotnet/arcade to build

Change build system to use arcade


lauxjpn:
It would also be useful to have a couple of different database environments setup and tested against, like in the `.travis.yaml` from `MySqlConnector`:

https://github.com/mysql-net/MySqlConnector/blob/0259c09fdef5bd5bd7dbb70c91093c48a3063afc/.travis.yml#L5-L26

caleblloyd:
Yes multiple DBs would be a good future addition, I think that's not too hard with the [matrix job strategy](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema#matrix)

lauxjpn:
Just an FYI:
The build paths have changed from their individual project directories to `/artifacts`.

lauxjpn:
This also touched a lot of files unnecessarily without changing any content. Is this a BOM issue, that should be handled for the whole repo?

Issue #839, opened 2019-09-27
Releasing 3.0.0-rc1.final


lauxjpn:
Let's use this issue to track the progress for a first ~~3.0.0 preview~~ final 3.0.0 rc1 release.

#797 is in use for general updates on the 3.0.0 GA progress so far.

lauxjpn:
Current test metrics:

MySQL | Total | Passed | Failed | Skipped
------|-------|--------|--------|--------
8.0.17 | 14104 | 13518 | 98 | 488
5.7.27 |  |  |  | 

(And yes, I did quickly fix a test before posting the stats, just to get under 100 for MySQL 8.)

CaddyDz:
Microsoft should seriously keep v2 of asp.net core in the homepage instead of v3 which every package I try to install now doesn't support yet

caleblloyd:
3.0.0 nightly builds are now available.  Feed instructions are updated on [README](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/README.md) and nightly builds are available on [Azure DevOps Feed](https://dev.azure.com/pomelo-efcore/Pomelo.EntityFrameworkCore.MySql/_packaging?_a=package&feed=pomelo-efcore-public&package=Pomelo.EntityFrameworkCore.MySql&version=3.0.0-rc1.19480.1&protocolType=NuGet&view=versions)

lauxjpn:
As far as I can see, the only serious issue blocking a preview release is #844.

lauxjpn:
I would like to (pre)release the `3.0.0-preview1` tomorrow, after the remaining open PRs have been merged.

Am I assuming correctly, that I just need to change the `PreReleaseVersionLabel` tag of the project's `Versions.prop` file and tag the merged commit accordingly to make the prerelease be published automatically to nuget?

caleblloyd:
I need to add the Nuget step for tags, working on that right now.

I think `3.0.0-rc1-final` is appropriate.  `-previewX` usually denotes it's tracking that version of EF Core Preview and that the API is still subject to change.  Since it's tracking EF Core RTM `rc` seems appropriate as we are trying to vet out bugs and find a stable release.

> Am I assuming correctly, that I would just need to change the PreReleaseVersionLabel tag of the project's Versions.prop file and tag the merged commit accordingly

[...]

PR #840, opened 2019-09-30
Fix remaining tests for MySQL 8.0.17

lauxjpn:
This fixes the remaining failing tests for MySQL 8.0.17.

To make them run, only very few minor changes to the actual code base were necessary. So I consider it pretty stable.

The following test class, that was previously enabled (but did fail partially) is being disabled at the moment:

- `MigrationsMySqlTest`

The following new test classes, have been disabled due to not yet implemented (or not fully implemented) support in the provider:

- `TransactionInterceptionMySqlTest`
- `TransactionInterceptionWithDiagnosticsMySqlTest`
- `UdfDbFunctionMySqlTest`
- `SpatialMySqlTest`
- `SpatialQueryMySqlTest`

All of these should be tackled for `3.0.0-preview2`.

Test metrics:

MySQL | Total | Passed | Failed | Skipped
------|-------|--------|--------|--------
8.0.17 | 14356 | 13859 | 0 | 497
5.7.27 | 14350 | 13777 | 69 | 504
5.6.45 | 14350 | 13586 | 260 | 504

The fixed tests for 5.7.27 and 5.6.45 will follow shortly.

This PR is in preparation of `3.0.0-preview1`: #839


lauxjpn:
Getting a couple of the following errors in the CI test runs (working fine on my local machine).

Windows:
```
MySql.Data.MySqlClient.MySqlException : Access denied for user 'root'@'localhost' (using password: NO)
```

Linux:
```
MySql.Data.MySqlClient.MySqlException : Access denied for user 'root'@'172.17.0.1' (using password: NO)
```

Does anybody have any idea how this is happening?

caleblloyd:
I do get the error locally using the same setup as CI Server (docker MySQL with a password)

The test passes when run on it's own:
```
./dotnet-env.sh dotnet test -c Release --logger trx test/EFCore.MySql.FunctionalTests --filter "FullyQualifiedName=Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Query.AsyncFromSqlQueryMySqlTest.Include_does_not_close_user_opened_connection_for_empty_result"
Test run for /home/caleb/Projects/caleb/Pomelo.EntityFrameworkCore.Mysql/artifacts/bin/EFCore.MySql.FunctionalTests/Release/netcoreapp3.0/Pomelo.EntityFrameworkCore.MySql.FunctionalTests.dll(.NETCoreApp,Version=v3.0)
Microsoft (R) Test Execution Command Line Tool Version 16.3.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

A total of 1 test files matched the specified pattern.
Results File: /home/caleb/Projects/caleb/Pomelo.EntityFrameworkCore.Mysql/test/EFCore.MySql.FunctionalTests/TestResults/_prec_2019-09-30_10_03_03.trx
                                                                                                                                                                                                               
Test Run Successful.
Total tests: 1
     Passed: 1
 Total time: 9.5957 Seconds
```

However the same 22 tests always fail when run all together

```
Test Run Failed.
Total tests: 14350
     Passed: 13824
     Failed: 22
    Skipped: 504
```

Typically I would suspect a tainted pooled Connection or DBContext but having the SAME tests fail every time does not support that.

Does not appear to be a MySqlConnector pooling issue as I tried Connection String with `Pooling=false` and `Max Pool Size=500` and the same 22 tests failed under each scenario.

Functional Tests do not appear to have DB Context Pooling enabled by default either.

So possibly something in a test fixture?

lauxjpn:
Thanks for replicating this issues on your end!
I assume it is some problem in the test code base then as well. I investigate this further on my end then.

lauxjpn:
The failing test are related to a connection that is using a password. For some reason, the password is being lost between the `Close()` and `Open()` calls so the connection cannot be opened again. All following tests of the same test class fail consequently.

I am trying to narrow down, if this is a Pomelo.MySql (test handling) or an MySqlConnector related issue.

lauxjpn:
It's getting weird. Now I am also loosing the password from the connection string while setting up tests in general, between ensuring the database exists and the call to clean the database.
It is being removed from the connection string during the `MySqlConnection.SetState()` call from this line:
https://github.com/mysql-net/MySqlConnector/blob/84a5288a8bba39d2adbbbbaefc2fec697df44b84/src/MySqlConnector/MySql.Data.MySqlClient/MySqlConnection.cs#L313

@bgrainger Is it expected, that ADO.NET is removing the password from the connection string in some instances (interestingly not every time) for security purposes?
If so, we need to adjust our test suit to not rely on the already set connection string of a connection.

lauxjpn:
The problem with setting up tests in general and the concrete failing test, seem to be caused by the same underlying behavior.

If a connection is opened, *might* loose it's password from the connection string. Because we reuse the connection string from a previous connection in the test suite, we *are* loosing the password at some point and further connections cannot be established.

bgrainger:
Yes, the password will be stripped from the connection string (once the connection is opened) if `Persist Security Info` is `false` (the default). See https://mysqlconnector.net/connection-options/ and https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnectionstringbuilder.persistsecurityinfo?view=netframework-4.8.

This shouldn't stop the connection being closed and opened again (if all you're doing is calling `Close()` and `Open()`). What you can't do is copy the connection string from an open `MySqlConnection` and use it to open a new connection.

> But it would be odd, that this is suddenly happening now, while this must have been working before.

This behaviour has been in MySqlConnector for a long time.

lauxjpn:
The problem with setting up tests in general and loosing the password in between seems to be expected by us, because we just handle it up the stack. So that was an hour well spend ;)

lauxjpn:
> Yes, the password will be stripped from the connection string (once the connection is opened) if Persist Security Info is false (the default).

That's it!

lauxjpn:
I will update the tests for 5.7 and 5.6 in a separate PR so other pending PRs can already be based on this one and profit from a working test suite in CI.

PR #848, opened 2019-10-03
Restore MySqlHistoryRepository to its pre-3.0.0-update state

### LINKED ISSUE #844: MySqlHistoryRepository is broken in 3.0
The current implementation of `MySqlHistoryRepository` was altered between 2.2.6 and 3.0.0.

By the looks of it, it should not work, because it currently uses statements like `IF`, which only work inside of stored procedures.

This needs to be checked and is currently blocking the 3.0.0-preview.

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/5675ac4b86f57622a62bac894c97b9dbdede8fea/src/EFCore.MySql/Migrations/Internal/MySqlHistoryRepository.cs#L66-L86

When fixing this, #776 and #810 should be considered.

lauxjpn:
The current 3.0.0 implementation was obviously "inspired" by the SQL Server implementation and does not honor MySQL's limited control flow statement support.

Fixes #844 and unblocks the ~~3.0.0-preview1~~ 3.0.0-rc1-final release #839.

### DIFF
--- a/src/EFCore.MySql/Migrations/Internal/MySqlHistoryRepository.cs
+++ b/src/EFCore.MySql/Migrations/Internal/MySqlHistoryRepository.cs
@@ -58,73 +58,43 @@ namespace Pomelo.EntityFrameworkCore.MySql.Migrations.Internal
         }
 
         /// <summary>
-        ///     This is an internal API that supports the Entity Framework Core infrastructure and not subject to
-        ///     the same compatibility standards as public APIs. It may be changed or removed without notice in
-        ///     any release. You should only use it directly in your code with extreme caution and knowing that
-        ///     doing so can result in application failures when updating to a new Entity Framework Core release.
+        ///     Overridden by database providers to generate a SQL Script that will <c>BEGIN</c> a block
+        ///     of SQL if and only if the migration with the given identifier does not already exist in the history table.
         /// </summary>
-        public override string GetBeginIfNotExistsScript(string migrationId)
-        {
-            Check.NotEmpty(migrationId, nameof(migrationId));
-
-            var stringTypeMapping = Dependencies.TypeMappingSource.GetMapping(typeof(string));
-
-            return new StringBuilder()
-                //.AppendLine($"DROP PROCEDURE IF EXISTS {MigrationsScript};")
-                //.AppendLine(@"DELIMITER //")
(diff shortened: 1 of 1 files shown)


src/EFCore.MySql/Migrations/Internal/MySqlHistoryRepository.cs
```diff
@@ -58,73 +58,43 @@ public override string GetCreateIfNotExistsScript()
...
-                .Append(stringTypeMapping.GenerateSqlLiteral(migrationId))
-                .AppendLine(")")
-                .Append("BEGIN")
-                .ToString();
-        }
+        /// <param name="migrationId"> The migration identifier. </param>
+        /// <returns> The generated SQL. </returns>
+        public override string GetBeginIfNotExistsScript(string migrationId) => $@"
+DROP PROCEDURE IF EXISTS {MigrationsScript};
+DELIMITER //
```
bgrainger: Is this actually being executed against a database server? `DELIMITER //` is a client-side directive that's not handled by MySqlConnector (https://github.com/mysql-net/MySqlConnector/issues/645#issuecomment-495767223) so it will get passed up to the server, where it will be treated as a syntax error.

lauxjpn: AFAIK this is only been used to generate a SQL script file, when the `--idempotent` argument is used in conjunction with `dotnet ef migrations script`.
See https://docs.microsoft.com/en-us/ef/core/miscellaneous/cli/dotnet for further details.

bgrainger: OK, I see. If they executed it in MySQL Workbench, it would be successful (because it understands the `DELIMITER` syntax).

lauxjpn: Yeah. It seems this was the consensus for the implementation (it is the unaltered code of 2.2.0).
It looks like the mysql command line tool supports this client-side syntax as well, so I think we are good for now:

https://dev.mysql.com/doc/refman/8.0/en/stored-programs-defining.html
https://dev.mysql.com/doc/refman/8.0/en/mysql-commands.html

PR #850, opened 2019-10-04
Add Matrix Test Strategy and Nuget Push on Tag

caleblloyd:
Adds matrix test strategy with:
- MySQL 8.0
- MySQL 5.7
- ~~MySQL 5.6~~
- MariaDB 10.4
- MariaDB 10.3

Adds Nuget Push on Tag


caleblloyd:
@lauxjpn I believe this is ready for review

What do you think about the MySql / MariaDB versions?  Are these the ones we want to support or do we want to drop 5.6?

Note that as configured, it will require CI passing on all builds before publishing release to Nuget.  `-wip` branches will still publish nightlies with CI failing.

test/EFCore.MySql.IntegrationTests/Tests/Attributes/SkipJsonFactAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public SkipJsonFactAttribute()
+        {
+            if (!AppConfig.ServerVersion.SupportsJson)
+            {
+                Skip = "Version of MySQL/MariaDB does not support JSON";
+            }
+        }
+    }
+}
```
lauxjpn: This can be removed, if the `SupportedServerVersionFactAttribute` is used instead.

test/EFCore.MySql.IntegrationTests/Tests/Models/GeneratedTypesTest.cs
```diff
@@ -12,7 +12,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.IntegrationTests.Tests.Models
 {
 	public class GeneratedTypesTest
 	{
-		[SkipAppVeyorAndMariaDbFact]
+		[SkipJsonFact]
```
lauxjpn: `SupportedServerVersionFactAttribute` can be used here like one of the two following lines:
```c#
[SupportedServerVersionFactAttribute(ServerVersion.JsonSupportKey)]
// or
[SupportedServerVersionFactAttribute(ServerVersion.JsonSupportKey, Skip = "Version of MySQL/MariaDB does not support JSON")]
```

test/EFCore.MySql.IntegrationTests/Tests/Attributes/SkipJsonFactAttribute.cs
caleblloyd: Ported `SupportedServerVersionFactAttribute` to `Pomelo.EntityFrameworkCore.MySql.IntegrationTests`

test/EFCore.MySql.IntegrationTests/Tests/Models/GeneratedTypesTest.cs
caleblloyd: Changed to `[SupportedServerVersionFact(ServerVersion.JsonSupportKey, Skip = "Version of MySQL/MariaDB does not support JSON")]`

caleblloyd:
How about supporting 2 minor MySQL and MariaDB Versions?  Seems like a reasonable maintenance policy, and if someone wanted to contribute support/tests for an older version we could accept as a PR.  Updated matrix and README with:

## Supported DBMS and Versions

`Pomelo.EntityFrameworkCore.MySql` is tested against the latest 2 minor versions of `MySQL` and `MariaDB`.  Older versions _may_ be compatible but are not officially supported or tested.  Currently supported versions are:

- MySQL 8.0
- MySQL 5.7
- MariaDB 10.4
- MariaDB 10.3

Issue #842, opened 2019-10-01
3.0: RenameColumn - SQL syntax error with MariaDB


PawelGerr:
A `MySqlException` is thrown when trying to rename a column using an EF migration.

### Migration command
```
migrationBuilder.RenameColumn(
                name: "OldName",
                table: "MyTable",
                newName: "NewName");
```
### Generated SQL statement
```
// 3.0.0-rc1.19480.1 - throws
ALTER TABLE `MyTable` RENAME COLUMN `OldName` `NewName`

// 2.2 - works
ALTER TABLE `MyTable` CHANGE `OldName` `NewName` bit NOT NULL DEFAULT FALSE;
```

### Exception
```
MySql.Data.MySqlClient.MySqlException (0x80004005): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'COLUMN `OldName` `NewName`' at line 1
   at MySqlConnector.Core.ServerSession.ReceiveReplyAsyncAwaited(ValueTask`1 task) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ServerSession.cs:line 772
```
Corresponding logger output
```
[15:11:38 ERR] [Microsoft.EntityFrameworkCore.Database.Command] Failed executing DbCommand (1,270,579ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
ALTER TABLE `MyTable` RENAME COLUMN `OldName` `NewName`
```

### Further technical details
MariaDB version: 10.4.7
Operating system: Win10-x64
Pomelo.EntityFrameworkCore.MySql version: 3.0.0-rc1.19480.1

lauxjpn:
The MariaDB support has not been tested very thoroughly yet, so we appreciate any reports related to that.

The new `RENAME COLUMN` syntax should only be used by the provider on MySQL >= 8.0.0.
Did you explicitly set the `ServerVersion` for the `DbContext`? If not, the provider currently assumes MySQL 8.0.0.
This is likely to change before the 3.0.0 release, where the server version will be detected automatically.
See #832

PawelGerr:
Nope, the version was missing but now I added it to the options builder:

```
builder
   .UseMySql(connectionString,
                     optionsBuilder => optionsBuilder.ServerVersion(
                                                            new Version(10,4,7),
                                                            ServerType.MariaDb))
```
Alas, I get the same error.

[...]

PR #852, opened 2019-10-04
Implement ServerVersion auto detection feature

### LINKED ISSUE #832: Expected MySqlOptions.ServerVersion behavior
What is the expected behavior, if `MySqlOptions.ServerVersion` does not get set at all?
Should we assume a default version and database type or should we throw an exception, with a clear message to set it up first before creating a `DbContext` or so?

There might also be the need for a convenient `AutoDetectServerVersion` method (or specially from `ServerVersion` derived class), that queries `@@version` or `VERSION()`, before the first query gets translated (or maybe just when a database connection gets opened).

lauxjpn:
The feature is discussed in #832, #156 and https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/842#issuecomment-537097399

This PR introduces the `MySqlConnectionInfo` class that implements the `IMySqlConnectionInfo` interface and contains the property `ServerVersion`.

If the `ServerVersion` is retrieved before the connection was opened (possible for some tests), it returns the default `ServerVersion` of `IMySqlOptions` with the `IsDefault` property set to `true`.
For the corresponding test cases it still needs to be evaluated, whether this behavior is expected or could potentially lead to tests for mixed server versions.

This *shouldn't* be an issue, as the test suit requires an explicit `ServerVersion` being set anyway, to handle the custom `Fact` and `Theory` attributes that depend on the `ServerVersion` from `AppConfig`.
So the auto detection feature would not be made use of for the test suite.

Fixes #832

### DIFF
--- a/src/EFCore.MySql/Design/Internal/MySqlDesignTimeServices.cs
+++ b/src/EFCore.MySql/Design/Internal/MySqlDesignTimeServices.cs
@@ -25,7 +25,8 @@ namespace Pomelo.EntityFrameworkCore.MySql.Design.Internal
                 .AddSingleton<IDatabaseModelFactory, MySqlDatabaseModelFactory>()
                 .AddSingleton<IProviderConfigurationCodeGenerator, MySqlCodeGenerator>()
                 .AddSingleton<IAnnotationCodeGenerator, MySqlAnnotationCodeGenerator>()
-                .TryAddSingleton<IMySqlOptions, MySqlOptions>();
+                .AddSingleton<IMySqlOptions, MySqlOptions>()
+                .AddSingleton<IMySqlConnectionInfo, MySqlConnectionInfo>();
         }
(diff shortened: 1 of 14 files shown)


lauxjpn:
> If the ServerVersion is retrieved before the connection was opened (possible for some tests), it returns the default ServerVersion of IMySqlOptions with the IsDefault property set to true.
For the corresponding test cases it still needs to be evaluated, whether this behavior is expected or could potentially lead to tests for mixed server versions.

Looks actually good to me as stated.

> This *shouldn't* be an issue, as the test suit requires an explicit ServerVersion being set anyway, to handle the custom Fact and Theory attributes that depend on the ServerVersion from AppConfig.
So the auto detection feature would not be made use of for the test suite.

In my tests, it *is not* an issue.

PR #853, opened 2019-10-05
Merge 2.2.6

caleblloyd:
PR form merging 2.2.6 to `master` and publishing a `rc` and then stable release

If we follow the trend of only maintaining the latest stable release, this will serve as the final 2.2.x stable release before 3.x gets merged to `master`

Things that still need to happen:
- [x] Upgrade MySqlConnector to latest version
- [x] Add CI code to publish package
- [x] Resolve merge conflicts


src/EFCore.MySql/EFCore.MySql.csproj
```diff
@@ -17,6 +17,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Relational" Version="$(MicrosoftEntityFrameworkCoreRelationalVersion)" />
     <PackageReference Include="MySqlConnector" Version="$(MySqlConnectorVersion)" />
     <PackageReference Include="Pomelo.JsonObject" Version="$(PomeloJsonObjectVersion)" />
+    <PackageReference Include="Castle.Core" Version="$(CastleCoreVersion)" />
```
bgrainger: I see that `Castle.Core` was added as a dependency between 2.2.1 and 2.2.6; is this necessary?

bgrainger: Added a comment here, where the dependency was added: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/commit/74bf3a90728a2c7998a4ff780d0d97e65a7e3be2#r35379779

lauxjpn: I totally missed that!

lauxjpn:
Removing the reference should be fine. If it really fails a test again, we can investigate this further.

PR #856, opened 2019-10-06
Fix non-deterministic MySQL errors due to dependent test classes

lauxjpn:
See https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/839#issuecomment-538706468


bgrainger:
:joy: at commit messages

lauxjpn:
See https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/839#issuecomment-538773518 and https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/839#issuecomment-538774906 for the reason and workaround.

lauxjpn:
I leave the verbosity and the `my.ini` in for the time being, though we might not need it in the end.

PR #857, opened 2019-10-06
Test PR Security

caleblloyd:
PRs should not be able to push to nuget feeds


caleblloyd:
Looks like AZDO secures service connections on PRs by default :+1: 

image

PR #859, opened 2019-10-06
Remove dependency on Castle.Core

bgrainger:
This appears to be unused: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/commit/74bf3a90728a2c7998a4ff780d0d97e65a7e3be2#r35379779

This removes an extra dependency that has been added between 2.2.1 and 2.2.6.


lauxjpn:
I think it was needed at the time to make a single test stop from failing. I can't remember which one it was though and even if it still fails, it might be enough then to just add `Castle.Core` to the test project instead.

PR #883, opened 2019-10-18
Extract primary key related SQL functions into migration header 2

igrmk:
This is the exact copy of https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/810
Somehow it doesn't work in CI due to cloning error


lauxjpn:
@igrmk Interesting, like last time, now my own PRs don't trigger CI either. I am not sure, whether this is connected somehow, or just a weird coincidence due to some AZDO problem or resource shortage.
/cc @caleblloyd

lauxjpn:
AZDO builds work again, though we did not change anything (AFAIK).

PR #886, opened 2019-10-19
Fixes #884 - Included null check for clrType when trying to match mappingInfo against Guid store types

### LINKED ISSUE #884: 3.0.0-rc1.final / BINARY(16) not mapped to Guid with OldGuids=true
### Steps to reproduce

**SQL**
```
CREATE DATABASE GuidDataTypeTest;
USE GuidDataTypeTest;

CREATE TABLE GuidTestTable (
	GuidTestTableID BIGINT NOT NULL AUTO_INCREMENT PRIMARY KEY,
	GuidTest1 BINARY(16) NOT NULL,
	GuidTest2 BINARY(16),
	GuidTest3 CHAR(36) NOT NULL,
	GuidTest4 CHAR(36)
);
```
**CLI**
`dotnet ef dbcontext scaffold "Server=localhost;Database=GuidDataTypeTest;User=user;Password=pw;TreatTinyAsBoolean=true;OldGuids=true" "Pomelo.EntityFrameworkCore.MySql"`

### The issue
Hi,

it seems that data type mapping for binary(16) (as well as CHAR(36)) is not working properly when scaffolding from database. BINARY(16) columns are mapped to byte[] and CHAR(36) to string when "OldGuids=true" is specified in connection string (but also when its omitted or set to false explicitly).

Generated entity code looks like this:
```
public partial class GuidTestTable
{
    public long GuidTestTableId { get; set; }
    public byte[] GuidTest1 { get; set; }
    public byte[] GuidTest2 { get; set; }
    public string GuidTest3 { get; set; }
    public string GuidTest4 { get; set; }
}
```

Additionally TINYINT(1) is not mapped to bool. I found that this has been explained / resolved in #866 and #875 - now I'm wondering if the mechanics described in there could also be responsible for this data type mapping issue. Did anyone else experience the same problem and already found a solution?

### Further technical details

MySQL version: 8.0.17
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 3.0.0-rc1.final
Microsoft.AspNetCore.App version: 3.0.0

[EDIT]
Just wanted to add that TINYINT(1) columns are mapped to bool correctly on nightly builds - BINARY916)/CHAR(36) still aren't.

henrypiassek:
BINARY(16) and CHAR(36) were not mapped to Guid when scaffolding from database due to clrType being null.

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
+++ b/src/EFCore.MySql/Storage/Internal/MySqlTypeMappingSource.cs
@@ -288,7 +288,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.Storage.Internal
                 if (_options.ConnectionSettings.OldGuids)
                 {
                     if (storeTypeName.Equals(_oldGuid.StoreType, StringComparison.OrdinalIgnoreCase)
-                        && clrType == typeof(Guid))
+                        && (clrType == typeof(Guid) || clrType == null))
                     {
                         return _oldGuid;
                     }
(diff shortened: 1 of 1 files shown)


lauxjpn:
@henrypiassek Thank you for your contribution!

henrypiassek:
@lauxjpn Well, thanks a lot for looking into it immediately and pointing out the issue right away - helped me a lot that day!

PR #903, opened 2019-10-30
Reimplement charset handling

### LINKED ISSUE #894: NCHAR, NVARCHAR and UnicodeCharSet
Currently `MySqlDbContextOptionsBuilder` contains an `AnsiCharSet` and a `UnicodeCharSet` extension method.
`MySqlTypeMappingSource` assumes Unicode for `nchar` and `nvarchar` types.

There might be some issues with that.

First of all we define the default `UnicodeCharSet` as `utf8mb4`:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/4740cb087ffc7045df8742ac1d4d4392a4b77bac/src/EFCore.MySql/Internal/MySqlOptions.cs#L16-L23

The official [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/charset-national.html) state the following however:

> Standard SQL defines NCHAR or NATIONAL CHAR as a way to indicate that a CHAR column should use some predefined character set. MySQL uses **utf8** as this predefined character set.

So the default for `nchar` and `nvarchar` should be `utf8`.

Also this charset seems to be predefined. So why do we provide the `UnicodeCharSet` extension method at all, if it cannot be changed?
I propose that we remove it.

Consequently we can rename `AnsiCharSet` to `CharSet`.

If no explicit charset is being specified, we should assume the default one for the used MySQL version, which is `latin1` < MySQL 8.0 >= `utf8mb4`.

Does anybody know, if MariaDB has made the move to `utf8mb4` as the default charset yet (and if so, in which version)?

lauxjpn:
- Removes `UnicodeCharSet` and (internally) introduces the hard-coded `NationalCharSet` (different behavior).
- Renames `AnsiCharSet` to `CharSet`.
- Implements all common MySQL charsets as static properties of the `CharSet` class. The class itself can be used to define custom charsets.
- From now on we use `utf8mb4` as the default charset for all character based store types except for `nchar` and `nvarchar` (which use `utf8`/`utf8mb3` predefined by MySQL and should be avoided as this is a proprietary UTF8 implementation).
- From now on we use `RelationalTypeMappingInfo.IsUnicode` and `MySqlStringTypeMapping([...] bool unicode, [...])` for actually trying to state that a type mapping uses Unicode, instead of stating that a column uses a national character store type.
- Fixes a design time service issue, where a previously initialized `MySqlOptions` instance could be unexpectedly changed.
- Assumes Unicode for requested type mappings with `IsUnicode` set to `null`, if the underlying charset has a max. byte per char count of >= 2.

To set the default charset back to the previous default of `latin1`, the following code can be used:

```c#
options.UseMySql(myConnectionString, b => b.CharSet(CharSet.Latin1));
```

Fixes #894


lauxjpn:
I added the following:

- Add `HasCharSet()` extension method for the `PropertyBuilder` class, which can be used to keep track of a database columns charset.
- Reverse Engineering/Scaffolding (Database --> Model) is supported as well as Migrations (Model --> Database).

I consider this PR complete now, so feel free to discuss and review.

Issue #922, opened 2019-11-07
Error applying migration: 


nuno-tiago-reis:
### Steps to reproduce

```
migrationBuilder.CreateTable
(
	name: "AspNetOrganizationRelations",
	schema: "Identity",
	columns: table => new
	{
		Id = table.Column<long>(nullable: false),
		Name = table.Column<string>(maxLength: 100, nullable: false),
		NormalizedName = table.Column<string>(maxLength: 100, nullable: false),
		CreatedAt = table.Column<DateTime>(nullable: false),
		UpdatedAt = table.Column<DateTime>(nullable: true)
	},
	constraints: table =>
	{
		table.PrimaryKey("PK_AspNetOrganizationRelations", x => x.Id);
	}
);

migrationBuilder.DropPrimaryKey
(
	name: "PK_AspNetOrganizationRelations",
	schema: "Identity",
	table: "AspNetOrganizationRelations"
);
```

### The issue
I'm applying a set of migrations that were working previously but now after upgrading to dotnet core 3.0 (and Pomelo.EntityFrameworkCore.MySql 3.0.0-rc2.final)  I'm getting the following error: 

MySql.Data.MySqlClient.MySqlException (0x80004005): PROCEDURE identity.POMELO_BEFORE_DROP_PRIMARY_KEY does not exist

```
fail: Microsoft.EntityFrameworkCore.Database.Command[20102]
      Failed executing DbCommand (7ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
      CALL POMELO_BEFORE_DROP_PRIMARY_KEY('Identity', 'AspNetOrganizationRelations');
      ALTER TABLE `Identity`.`AspNetOrganizationRelations` DROP PRIMARY KEY;
Application startup exception: MySql.Data.MySqlClient.MySqlException (0x80004005): PROCEDURE identity.POMELO_BEFORE_DROP_PRIMARY_KEY does not exist
 ---> MySql.Data.MySqlClient.MySqlException (0x80004005): PROCEDURE identity.POMELO_BEFORE_DROP_PRIMARY_KEY does not exist
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior) in C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs:line 49
[...]

Issue #897, opened 2019-10-24
Multi-schema drop


Donistivanov:
### Steps to reproduce
I have a MySQL conn string: `server=localhost;User Id=groot;password=password;database=schema1`

I put a Table attribute on some model like: `[Table("cool_table", Schema="schema2")]`

Scaffolding and db update makes both schemas.

`dotnet ef database drop` only drops schema1

### The issue
I would like to be able to drop all the schemas related to my project

### Further technical details

MySQL version: 6.3
Operating system: Windows
Pomelo.EntityFrameworkCore.MySql version: `<PackageReference Include="Pomelo.EntityFrameworkCore.MySql" Version="2.2.0-preview1-final" />`
Microsoft.AspNetCore.App version: `<TargetFramework>netcoreapp2.2</TargetFramework>`

Other details about my project setup: If I have to have references between schemas, I think I had to put the line `modelBuilder.HasDefaultSchema("schema1");`

mguinness:
Similar to https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/892#issuecomment-546073417 you'll have to create a `DbContext` for that database (if you haven't already) and specify that in the [CLI](https://www.entityframeworktutorial.net/efcore/cli-commands-for-ef-core-migration.aspx#database-drop).

Issue #892, opened 2019-10-24
DatabaseFacade.EnsureCreated throws 'MySqlDbException'


LeaFrock:
### Steps to reproduce
`using var context = new MyDbContext(DbConfig);`
`context.Database.EnsureCreated();`

As the annotation says, the method should return a 'bool' value instead of throwing an exception. But in fact it throws and the message tells 'Table xxxx already exists'.
So I turn on the 'general_log' of local MySql, check the sql it gets and I find something wrong.  Please see the image below.
![20191024181014](https://user-images.githubusercontent.com/20787265/67475895-9870ea00-f689-11e9-8465-1a9e26bbfe25.png)

### The issue
I wonder what's wrong about the problem above. Why doesn't the sql of table creating contain '**if not exists**'? Is it a bug or just my misunderstanding?

Exception message:
`MysqlDbException: Table MyTableName already exists`
Stack trace:
   at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet() in C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs:line 125
   at MySql.Data.MySqlClient.MySqlDataReader.<CreateAsync>d__95.MoveNext() in C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs:line 338
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySqlConnector.Core.CommandExecutor.<ExecuteReaderAsync>d__0.MoveNext() in C:\projects\mysqlconnector\src\MySqlConnector\Core\CommandExecutor.cs:line 63
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #923, opened 2019-11-07
Throw on schema use

lauxjpn:
Throw if a non-empty schema has been set, as MySQL does not support the EF Core concept of schemas.

This might be a breaking change for anybody who used that loophole in previous versions.

Addresses #922, #897, #892 and possibly others.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -110,16 +110,16 @@ jobs:
     displayName: Integration Tests legacy migrations
     continueOnError: true
     condition: eq(variables['INTEGRATION_TESTS'],'true')
-  - bash: export EF_SCHEMA="pomelo_test2" && ./dotnet-env.sh dotnet build ./test/EFCore.MySql.IntegrationTests -c Release
+  - bash: export EF_DATABASE="pomelo_test2" && ./dotnet-env.sh dotnet build ./test/EFCore.MySql.IntegrationTests -c Release
     displayName: Integration Tests Building migrations with EF_DATABASE=pomelo_test2
     continueOnError: true
     condition: eq(variables['INTEGRATION_TESTS'],'true')
-  - bash: export EF_SCHEMA="pomelo_test2" && ./dotnet-env.sh ./test/EFCore.MySql.IntegrationTests/scripts/rebuild.sh
+  - bash: export EF_DATABASE="pomelo_test2" && ./dotnet-env.sh ./test/EFCore.MySql.IntegrationTests/scripts/rebuild.sh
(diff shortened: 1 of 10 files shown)


lauxjpn:
Just as an FYI, if anybody really needs to reactivate this broken functionality (which i highly discourage), you could derive a class from `MySqlMigrationsSqlGenerator`, override `CheckSchema(MigrationOperation operation)` to not throw and finally inject this custom `IMigrationsSqlGenerator` implementation in the DI container.

This is clearly *unsupported*, but currently possible, as all the previous schema mechanics are still in place.

jgcjoec:
New to github so please forgive me if I am violating any rules. I recognize that the team is identifying this as a breaking change, which is great, but I'm questioning . We have an app that supports multiple RDBMSes, each with different capabilities. Some supported RDBMSes support "schemas", some don't. 

To continue to support multiple RDBMSes, we now need to implement an unsupported workaround, which doesn't seem like a good solution, or we need to somehow add logic to our app to detect which RDBMS we're hooked to at runtime and conditionally add schemas. Why couldn't an info/warning be displayed at runtime instead? I looked at the linked issues above and I understand that this exception was added because "users trip over it a lot", but wouldn't a info/warning to STDOUT/STDERR solve the problem with less friction?

Thanks,

- Joe

lauxjpn:
This was always a hack and is not implemented in other providers like Sqlite as well. It leads to weird and inconsistent behavior in many cases, especially when it come to database generation or deletion.

It is currently discussed in the EF Core team, if they will move more to a DBMS specific migration model, as the current one leads to other problems and limitations.

The [Migrations with Multiple Providers](https://docs.microsoft.com/en-us/ef/core/managing-schemas/migrations/providers?tabs=dotnet-core-cli) MSDN article does provide options for using multiple providers in general.

I did provide a workaround in the comment above, that makes it currently possible for anybody to activate the previous behavior, though this is probably not going to work anymore in a year with .NET 5.
Instead, it would be best to upgrade your code now.

There are multiple ways to do this and if you need any help, just open a new issue and post us some example code and what you expect and we take a look at it.

jgcjoec:
@lauxjpn thanks for the quick response! I discussed your options listed above with my team here and we prefer the first option - "We could just ignore any supplied schema but not throw".

I personally think some kind of run-time warning, when a schema is supplied (which maybe can be suppressed by a developer explicitly in code, if they choose?), might give the best of both worlds.

lauxjpn:
I opened #972 to continue the discussion about this.

Issue #913, opened 2019-11-05
Scaffold HasCharSet generation for tables


mguinness:
Investigate omitting `HasCharSet()` when the column charset matches the table, to align with [Column Character Set and Collation](https://dev.mysql.com/doc/refman/8.0/en/charset-column.html).

We would have to annotate the entity with the table's charset (like we do with properties and their columns now) to support that.

lauxjpn:
Additionally, we can think about making the `HasCharSet()` scaffolding optional. For example, a connection string pseudo-option like `Scaffold:CharSets=true` could be introduced, with a reasonable default value.

Pomelo and MySqlConnector are using `System.String` internally and MySqlConnector is using `utf8mb4` for all its server communication on the wire.

So the `HasCharSet()` calls really only matter, if the migration code needs to create tables with the exact same characteristics of the source tables it was created of. That should only be needed, if external dependencies on the database exists (e.g. other apps).

For those scenarios, supporting collations would be the next step, as these are not used by Pomelo either, while external apps could depend on them.

mguinness:
It looks like there is a discussion about `RelationalTypeMapping` in https://github.com/aspnet/EntityFrameworkCore/issues/11896 so maybe @roji can give some input.

What would be neat is if charset is `latin1` that the following data annotation could be added to the model validation to avoid data loss.

```C#
[RegularExpression(@"^[\x20-\x7E\xA0-\xFF]+$", ErrorMessage = "Only ANSI characters accepted")]
public string Name { get; set; }
```

lauxjpn:
Only scaffolding the column charset if it is different from the tables default charset might lead to unexpected behavior later on.

This is due to the fact, that if no charset is defined for a column at creation time, MySQL creates the column with the tables default charset. If at a point later in time, the tables default charset is changed, the charset of the column that was created using the tables default charset will stay unchanged.

Let's assume the following initial model definition:

```c#
modelBuilder.Entity<MyEntity>(entity =>
    entity.HasCharSet("latin1");

    entity.Property(e => e.TextProperty)
        .HasMaxLength(255);
});
```

When this model is created, `TextProperty` will have a charset of `latin1`.
We later change the default charset:

```c#
modelBuilder.Entity<MyEntity>(entity =>
    entity.HasCharSet("utf8mb4");

    entity.Property(e => e.TextProperty)
        .HasMaxLength(255);
});
```

[...]

PR #931, opened 2019-11-10
Introduce scaffolding specific connection string options

lauxjpn:
These settings can be used to control what the scaffolder generates.
Currently the following options are supported:
- `Scaffold:CharSet` (default is `true`) that generates `.HasCharSet()` calls for character based columns if `true`
- `Scaffold:Collation` (default is `true`) that generates `.HasCollation()` calls for character based columns if `true`

I wasn't sure if singular or plural option names are more intuitive, so I just went with singular ones.

A scaffolding command could look like this:

`dotnet ef dbcontext scaffold "server=127.0.0.1; uid=root; pwd=; port=3306; database=MyDatabase; Scaffold:CharSet=False; Scaffold:Collation=True"`

Addresses https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/913#issuecomment-550557616

### DIFF
--- a/src/EFCore.MySql/Migrations/Internal/MySqlMigrator.cs
+++ b/src/EFCore.MySql/Migrations/Internal/MySqlMigrator.cs
@@ -52,6 +52,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.Migrations.Internal
             bool idempotent = false)
         {
             IEnumerable<string> appliedMigrations;
+
             if (string.IsNullOrEmpty(fromMigration)
                 || fromMigration == Migration.InitialDatabase)
             {
@@ -76,11 +77,13 @@ namespace Pomelo.EntityFrameworkCore.MySql.Migrations.Internal
             var dropPrimaryKeyExists = operations.Any(x => x is DropPrimaryKeyOperation);
             var addPrimaryKeyExists = operations.Any(x => x is AddPrimaryKeyOperation);
 
-            List<string> parts = new List<string>();
+            var parts = new List<string>();
+
             if (dropPrimaryKeyExists)
             {
                 parts.Add(PrepareString(BeforeDropPrimaryKeyHeader));
             }
+
             if (addPrimaryKeyExists)
             {
                 parts.Add(PrepareString(AfterAddPrimaryKeyHeader));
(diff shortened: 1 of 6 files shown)


lauxjpn:
Quick question at @AndriySvyryd or @roji:
Am i correct in assuming, that `The implementation does not need to be thread-safe.` for a `MigrationsSqlGenerator` derived class like our `MySqlMigrationsSqlGenerator` means, that there should be no unexpected side effects from temporarily saving a state like the current `IModel`, so we can use it in methods like `DefaultValue()` that otherwise do not have access to this information due to lack of a corresponding parameter?

roji:
/cc @bricelam who is the migrations guru.

To me that sounds like it *should* be safe - as long as you make sure it's always set by an upper method in the stack and then read by the lower one. In other words, take care that the function looking at the model never sees a stale one from a previous operation or something.

However, if you really need the model in DefaultValue you may be better off simply overriding ColumnDefinition instead and doing your thing there (also override DefaultValue to throw in case it gets called from somewhere else in the future). In any case, could you share what stuff you need from the model at that point? We may want to consider improving the APIs in this area...

lauxjpn:
Thanks for the quick input!

This is just about a compatibility issue, where the syntax for default values has been extended for MySQL 8.0.13 and higher versions to support expressions. A side effect of this is, that it will allow us to support default values for some types (e.g. `TEXT` and `BLOB`) that could not be used in conjunction with default values before.

This question actually better belongs to the #932 PR, were we introduced an annotation and extension method to specify the server version that a model is intended for.

To fully make use of the intended server version, we might need access to it in any `MigrationsSqlGenerator` method, which means we need to temporarily store it while we process a `MigrationOperation`.

If it turns out (very unlikely), that enough users don't comply with the specs and use multi-threaded calls on the `MigrationsSqlGenerator` class and for some reason cannot sync their calls, we could still implement some sync mechanisms.

I just wanted to make sure, that there isn't anything fancy in the pipeline for EF Core and the strait forward and sequential processing of the `MigrationOperation` collection is likely to stay.

ajcvickers:
@lauxjpn We really appreciate all the work you're doing here! We would love to hear how things are going and get any feedback you have on how we could make things easier for you. Would it work to setup a conference call? (I see you're in Berlin, which is actually where @roji is based! (Although he is here in Redmond for a couple of weeks.))

You can email me at Microsoft if you're interested--I'm `avickers` there.

lauxjpn:
@ajcvickers Thanks, I am glad I can help. I am sure we can setup something next week or so. I'll send you an email about it tomorrow.

Also, our `3.0.0` support is complete an should go GA sometime around this weekend. We are just polishing up a few things and making sure nobody complains about any major issues since the last prereleases.

PR #932, opened 2019-11-10
Support server version specific models and migrations

lauxjpn:
Introduces the annotation `MySql:ServerVersion` and the `ModelBuilder` extension method `ForServerVersion()`.

The annotation is automatically been generated when scaffolding and will be part of migrations if specified in the model definition the migrations are generated from.

The server version that a model is intended for can be specified like in the following code (it can also be specified as a `ServerVersion` object instead of a string):

```c#
protected override void OnModelCreating(ModelBuilder modelBuilder)
{
    modelBuilder.ForServerVersion("5.7.27-mysql");

    // ...
}
```

Migrations will generate SQL scripts based on this server version, e.g. regarding a `RenameColumn` operation an `ALTER TABLE RENAME COLUMN` statement will be generated for a server version of `8.0.0-mysql` or higher, while an `ALTER TABLE CHANGE COLUMN` statement will be generated for versions below that.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -79,7 +79,7 @@ jobs:
   - bash: ./dotnet-env.sh dotnet test --logger trx test/EFCore.MySql.Tests
     displayName: Tests
     continueOnError: true
-  - bash: ./dotnet-env.sh dotnet test test/EFCore.MySql.FunctionalTests -c Release --logger trx --verbosity detailed
+  - bash: ./dotnet-env.sh dotnet test test/EFCore.MySql.FunctionalTests -c Release --logger trx
     displayName: Functional Tests
     continueOnError: true
   - bash: ./dotnet-env.sh dotnet run --project test/EFCore.MySql.IntegrationTests -c Release testMigrate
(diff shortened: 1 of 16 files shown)


AndriySvyryd:
Why is this change needed? With the current approach of specifying the server version in the context options the migrations can be applied to different servers without modifying the model.

lauxjpn:
I think this would not work when generating scripts for example. Though it might be a good idea to add  the new annotation in addition the the current approach, so the model annotation is only used if no server version was setup (or auto detected).

AndriySvyryd:
Script generation should use the configured version as well. It seems to be a low value feature. And considering that it will need to be maintained it's probably best not to add it.

lauxjpn:
>Script generation should use the configured version as well.

You are actually correct. An explicit call to `MySqlDbContextOptionsBuilder.ServerVersion()` *will* be used for script generation (the auto detection feature will not work, as no database connection is being initiated) even when using the command line tools.

Consequently, annotation and extension method are redundant and should be abandoned.
Good catch!

>It seems to be a low value feature.

I think as it looks now, it is actually a *no value* feature, or is there any case this would still be useful for?

roji:
You may want to take a look at what Npgsql does for server versions ([you can start here](https://github.com/npgsql/efcore.pg/blob/a2180f03038086f9c92e60ab1d97ef567f3d76db/src/EFCore.PG/Infrastructure/NpgsqlDbContextOptionsBuilder.cs#L45)). As @AndriySvyryd suggests above, it's a context option rather than a model annotation, and can be easily used in any service that needs to do version-dependent logic. If you need any help understanding/implementing something like this, let me know.

lauxjpn:
@roji Thanks. We actually already got one, I just missed that it also works when just generating scripts and not only, when an actual database round trip is being made.

PR #969, opened 2019-11-26
Ensure that generated GUIDs conform to the RFC 4122 standard.

sgielen:
RFC 4122 describes the format of a UUID (or as Microsoft calls them, GUID) as
xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx, where M is the UUID version and some bits
of N denote the UUID variant. The GUID generator in this repository sets these
nibbles to random values, making the generated UUIDs incompatible with the RFC.

This may cause problems later on, as there is a chance future UUID
implementations may misbehave on nonconformant UUIDs or not accept them at all.
For example, UUID libraries may assume that a UUID with version 0 is a nil uuid
and compare all such UUIDs to be equal to 00000000-0000-0000-0000-000000000000
and each other.

This commit fixes this by setting the UUID version and variant properly in
generated UUIDs. The same number of bytes is still used for the ticks, while
there is now one byte less randomness in each UUID. However, together with the
ticks in 100 ns precision, the chance of a collision is still negligible.  The
version used is 4, for a UUID based on randomness; we hide the fact that most
of the UUID is not truly random but based on time. The variant is 0b10xx for
RFC 4122.

This commit also adds tests for the new properties as well as the old
properties presumed by the existing implementation, such as that generated
GUIDs are sorted on time and unique, non-empty GUIDs are generated.


src/EFCore.MySql/Storage/Internal/MySqlConnectionSettings.cs
```diff
@@ -35,7 +35,7 @@ public MySqlConnectionSettings(string connectionString)
             TreatTinyAsBoolean = csb.TreatTinyAsBoolean;
         }
 
-        public MySqlGuidFormat GuidFormat { get; }
+        public virtual MySqlGuidFormat GuidFormat { get; }
```
sgielen: This change is necessary to allow Moq to override the value returned from this property. Perhaps there are other ways to allow this, too.

mguinness:
Is possible to reuse Sequential GUIDs code in [SequentialGuidValueGenerator](https://github.com/aspnet/EntityFrameworkCore/blob/master/src/EFCore/ValueGeneration/SequentialGuidValueGenerator.cs) from EF Core instead?

mguinness:
@caleblloyd I saw that you did an earlier [Sequential GUIDs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/260) PR, do you have any comments?

lauxjpn:
Thank you for your PR!

Though I don't share your concerns about future incompatibilities of the current implementation, I think it is as good idea to advertising conformance with Variant 1, Version 4 of RFC 4122 and I don't see any downsides to your approach.

lauxjpn:
@mguinness Interestingly, SQL Server 14 does not seem to implement its `NEWSEQUENTIALID()` function anymore like `SequentialGuidValueGenerator`, but actually more like we do.

@caleblloyd's optimized implementation for both, binary based and character base columns, is still preserved with this PR. It's just being fixed to conform to the standard.

ajcvickers:
@smitpatel @bricelam @AndriySvyryd @roji @maumar Interesting discussion here on GUIDs. Most specifically, that SQL Server may have changed sequential GUID generation since we last looked at it. Should we react in EF?

caleblloyd:
It makes sense to conform to the RFC. Thanks for adding the tests. I think they need to be moved out of the Functional tests and into the Unit tests at `test/EFCore.MySql.Tests`. The unit test project already included Moq.

roji:
@ajcvickers yeah, we should definitely consider changing on our side...

ajcvickers:
@roji Can you file an issue outlining what you think we should do?

src/EFCore.MySql/Storage/Internal/MySqlConnectionSettings.cs
lauxjpn: While you are at it, please make `TreatTinyAsBoolean` virtual as well. We should have made then both virtual in the first place.

src/EFCore.MySql/ValueGeneration/Internal/MySqlSequentialGuidValueGenerator.cs
```diff
@@ -42,21 +65,24 @@ public override Guid Next(EntityEntry entry)
                     Array.Reverse(tickBytes);
                 }
 
-                Buffer.BlockCopy(tickBytes, 0, guidBytes, 0, 8);
-                Buffer.BlockCopy(randomBytes, 0, guidBytes, 8, 8);
+                Buffer.BlockCopy(tickBytes, 0, guidBytes, 0, 6);
+                guidBytes[6] = (byte)(ticksAndVersion << 8 >> 8);
+                guidBytes[7] = (byte)(ticksAndVersion >> 8);
+                guidBytes[8] = ticksAndVariant;
+                Buffer.BlockCopy(randomBytes, 0, guidBytes, 9, 7);
```
lauxjpn: Please change the rest of the `Buffer.BlockCopy` calls to simple assignments as well. There is probably not much advantage using them over assignments here anymore.

mguinness:
> Interestingly, SQL Server 14 does not seem to implement its NEWSEQUENTIALID() function anymore like SequentialGuidValueGenerator, but actually more like we do.

At a lower level `NEWSEQUENTIALID()` function uses Windows API `UuidCreateSequential` that uses MAC address (see [SQL Server on Linux: CU4 – NewSequentialId() – Uuid](https://bobsql.com/sql-server-on-linux-cu4-newsequentialid-uuid/)), which is different to `NewGuid()` used by [CoreFX in Unix](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Guid.Unix.cs).  It's not relevant to this discussion, but `NEWSEQUENTIALID()` is not RFC 4122 compliant.

lauxjpn:
@mguinness Good links! They also lead to [UNRAVELING THE MYSTERIES OF NEWSEQUENTIALID](http://www.jorriss.net/2008/04/24/unraveling-the-mysteries-of-newsequentialid/) with a practical analysis and examples of the differences between the individual `Guid` generation methods, including `NEWSEQUENTIALID()`.

According to the link, something similar to the following pseudo code is being done:

```
newGuid = UuidCreateSequential();
bytes = newGuid.ToByteArray()
Array.Reverse(bytes, 0, 4)
Array.Reverse(bytes, 4, 2)
Array.Reverse(bytes, 6, 2)
```

The proposed solution here (base on the previous work of @caleblloyd) is similar to `NEWSEQUENTIALID()` in its effect to generate 128 bit values with improved sorting/indexing capabilities, has a higher level of privacy due to not making use of MAC addresses and returns a RFC conforming `Guid`.

sgielen:
Thanks for your feedback, all! I've addressed relevant proposed changes in the last commit.

roji:
Issue on the EF Core repo to investigate behavior on the SQL Server side: https://github.com/aspnet/EntityFrameworkCore/issues/19124

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Linq;
+using Moq;
+using MySql.Data.MySqlClient;
+using Pomelo.EntityFrameworkCore.MySql.Internal;
+using Pomelo.EntityFrameworkCore.MySql.Storage.Internal;
+using Pomelo.EntityFrameworkCore.MySql.ValueGeneration.Internal;
+using Xunit;
+
+// ReSharper disable InconsistentNaming
+namespace Pomelo.EntityFrameworkCore.MySql.FunctionalTests.ValueGeneration.Internal
```
lauxjpn: The test class was moved to `EFCore.MySql.Tests`, so the namespace should be updated as well (see other test classes for reference).

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+{
+    public class MysqlSequentialGuidValueGeneratorTest
+    {
+        private MySqlSequentialGuidValueGenerator getGenerator(bool mysqlGuidFormatIsLittleEndianBinary16) {
```
lauxjpn: Use PascalCasing for methods (including private methods).

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+            connectionSettings.SetupGet(x => x.GuidFormat).Returns(mysqlGuidFormatIsLittleEndianBinary16 ? MySqlGuidFormat.LittleEndianBinary16 : MySqlGuidFormat.None);
+
+            var options = new Mock<MySqlOptions>();
+            options.SetupGet(x => x.ConnectionSettings).Returns(connectionSettings.Object);
+
+            return new MySqlSequentialGuidValueGenerator(options.Object);
+        }
+
+        [Fact]
+        public void NextReturnsNewGuid()
```
lauxjpn: Change name to `Next_returns_new_Guid`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var guid = getGenerator(false).Next();
+            Assert.NotEqual(Guid.Empty, guid);
+        }
+
+        [Fact]
+        public void NextReturnsNewGuidLEB16()
```
lauxjpn: Change name to `Next_returns_new_Guid_LittleEndianBinary16`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var guid = getGenerator(true).Next();
+            Assert.NotEqual(Guid.Empty, guid);
+        }
+
+        [Fact]
+        public void NextReturnsLooselySortedGuid()
```
lauxjpn: Change name to `Next_returns_loosely_sorted_Guid`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+            for(var i = 0; i < 100; ++i) {
+                timeNow = timeNow.AddTicks(1);
+                var guid = generator.Next(timeNow);
+                Assert.True(string.Compare(guid.ToString(), lastGuid.ToString()) > 0);
+                lastGuid = guid;
+            }
+        }
+
+        [Fact]
+        public void NextReturnsLooselySortedGuidLEB16()
```
lauxjpn: Change name to `Next_returns_loosely_sorted_Guid_LittleEndianBinary16`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+            for(var i = 0; i < 100; ++i) {
+                timeNow = timeNow.AddTicks(1);
+                var guid = generator.Next(timeNow);
+                Assert.True(string.Compare(guid.ToString(), lastGuid.ToString()) > 0);
+                lastGuid = guid;
+            }
+        }
+
+        [Fact]
+        public void NextReturnsValidRFC4122RandomGuid()
```
lauxjpn: Change name to `Next_returns_valid_Rfc4122_random_Guid`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+                var version = bytes[7] >> 4;
+                Assert.Equal(4, version);
+                // Variant, indicated by the first bits of the 17th nibble, must be 0b10xx
+                var variant = (bytes[8] >> 4) & 0b1100;
+                Assert.Equal(0b1000, variant);
+            }
+        }
+
+        [Fact]
+        public void NextReturnsValidRFC4122RandomGuidLEB16()
```
lauxjpn: Change name to `Next_returns_valid_Rfc4122_random_Guid_LittleEndianBinary16`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+                var version = bytes[7] >> 4;
+                Assert.Equal(4, version);
+                // Variant, indicated by the first bits of the 17th nibble, must be 0b10xx
+                var variant = (bytes[8] >> 4) & 0b1100;
+                Assert.Equal(0b1000, variant);
+            }
+        }
+
+        [Fact]
+        public void NextReturnsUniqueGuid()
```
lauxjpn: Change name to `Next_returns_unique_Guid`.

test/EFCore.MySql.Tests/ValueGeneration/Internal/MysqlSequentialGuidValueGeneratorTest.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var generator = getGenerator(false);
+
+            var guids = Enumerable.Range(0, 100).Select(_ => generator.Next()).ToArray();
+            Assert.Equal(guids.Distinct().Count(), guids.Length);
+        }
+
+        [Fact]
+        public void NextReturnsUniqueGuidLEB16()
```
lauxjpn: Change name to `Next_returns_unique_Guid_LittleEndianBinary16`.

sgielen:
Thank you @lauxjpn for your feedback, I've addressed the naming issues you mentioned.

lauxjpn:
@sgielen Thank you for your contribution!

PR #979, opened 2019-12-04
Override more tests that rely on implicit ordering behavior of SQL Server

lauxjpn:
Override more tests that rely on implicit ordering behavior of SQL Server and implement explicit sorting instead. Without this fix, these 3 tests would only fail on `8.0.18-mysql` on Linux (but worked on lower versions and also work on all version on Windows).

Fixes the current CI pipeline that uses the `8.0.18-mysql` version on Linux.

### DIFF
--- a/test/EFCore.MySql.FunctionalTests/Query/IncludeAsyncMySqlTest.cs
+++ b/test/EFCore.MySql.FunctionalTests/Query/IncludeAsyncMySqlTest.cs
@@ -1,4 +1,9 @@
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Query;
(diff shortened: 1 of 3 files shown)


ajcvickers:
@lauxjpn For cases of SQL Server-specific behavior we have also in the past taken PRs to the tests in the EF Core repo itself. This is often better than overriding like this because other providers will then not hit the SQL Server-specific issue. I believe @roji and @ErikEJ have both submitted such PRs in the past.

Of course, it's totally up to you--feel free to keep doing it this way if you prefer.

/cc @maumar @smitpatel

lauxjpn:
I actually have an item for opening an issue about similar tests I came across while doing the 3.0 upgrade on my TODO list. I will dig them out again in the second January half, when I got more time on my hands and plan to go over all of the tests again anyway. I will be happy to provide a PR then.

ajcvickers:
@lauxjpn Great! Thanks.

PR #982, opened 2019-12-04
Implement alternatives to the current behavior to always throw, if a schema has been set for an object

### LINKED ISSUE #972: Provide alternative options to throwing in case of schema use with multiple providers
This is a continuation of the discussion from #923 about alternatives to throwing, in case Pomelo is being used as one of many providers in conjunction with the "single migration for multiple providers" approach and other providers must use schemas.

To make this very clear:

This is **not** a discussion about reimplementing the previously dropped version, where schemas were misused as a way to access objects from multiple databases by a single `DbContext`.

It is about alternatives that are compliant with EF Core's understanding of schemas.

Proposed alternative options have been to either just ignore schemas (but not throw) or allow users to define their own translation function, that translates an input schema and an input object name to an output object name (within the same database).

Any option implemented here is going to be opt-in, because MySQL does not support the EF Core concept of schemas and therefore should throw by default if a schema is being used anyway. The exception should hint at the alternative options.

lauxjpn:
Threre are currently 3 options to choose from:

```c#
// Throw an exception, if a schema is being used. This is the default.
options.UseMySql(myConnectionString, b => b.SchemaBehavior(MySqlSchemaBehavior.Throw))

// Silently ignore any schema definitions.
options.UseMySql(myConnectionString, b => b.SchemaBehavior(MySqlSchemaBehavior.Ignore))

// Use the specified translator delegate to translate from an input schema and object name to
// an output object name whenever a schema is being used.
options.UseMySql(myConnectionString, b => b.SchemaBehavior(MySqlSchemaBehavior.Translate,
    (schema, entity) => $"{schema ?? "dbo"}_{entity}"))
```
This still needs some testing.

Fixes #972

### DIFF
--- a/src/EFCore.MySql/Infrastructure/Internal/IMySqlOptions.cs
+++ b/src/EFCore.MySql/Infrastructure/Internal/IMySqlOptions.cs
@@ -18,5 +18,6 @@ namespace Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal
         bool NoBackslashEscapes { get; }
         bool ReplaceLineBreaksWithCharFunction { get; }
         MySqlDefaultDataTypeMappings DefaultDataTypeMappings { get; }
+        MySqlSchemaNameTranslator SchemaNameTranslator { get; }
     }
 }
--- a/src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
+++ b/src/EFCore.MySql/Infrastructure/Internal/MySqlOptionsExtension.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Pomelo Foundation. All rights reserved.
 // Licensed under the MIT. See LICENSE in the project root for license information.
 
+using System;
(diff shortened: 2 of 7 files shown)


rubenruvalcaba:
HI,

I'm just facing the same problem, my application must run either on SQL Server or MySQL. The translator  solution will work for me, but in EF Pomelo.EntityFrameworkCore SchemaBehavior and MySqlSchemaBehavior are not available. Apparently it's only for NET Framework version. How can I achieve the same on EF Core?

lauxjpn:
@rubenruvalcaba Just use our [nightly build feed](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#nightly-builds). It contains the latest bits, which also includes this PR (see "Milestone" on the right sidebar on this page). The 3.2.0 version has not been officially released yet.

rubenruvalcaba:
Great! Thank you I'll try it out

Regards

PR #1017, opened 2020-02-03
Use case-insensitive comparison on table names when scaffolding foreign key constraints

dhirensham:
MySQL Workbench names tables in lower-case by default. If you capitalize your table names and then use Workbench to create foreign key constraints, the referenced table name stored in the information schema won't match the table name, and the navigation property won't be created.

This changes the comparison to a case-insensitive one.


lauxjpn:
CI will be fixed shortly.

The [9.2.3 Identifier Case Sensitivity](https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html) of the docs will be relevant to the discussion.

As for the actual issue, we should discuss under wich circumstances this problem you try to fix arises. What OS is your MySQL instance running under, which MySQL version are you using and what is the value of your [lower_case_table_names](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) status variable?

As far as I can see, this is only related to the MySQL Workbench specific option `SQL Identifiers are Case Sensitive` under `Edit -> Preferences -> General Editors -> SQL Parsing in Code Editors`. If the option is disabled, MySQL Workbench will use a lower case table name when adding foreign keys to a table.

I therefore assume, that you have/had the option disabled, correct?

This might be a bug in MySQL Workbench, because it should use the actual table name and not assume some casing when altering the table with added constraints. So this should also be reported on the MySQL Workbench bug tracker.

I think this PR needs some alterations. Using a case sensitive comparison should still be the primary mechanism to correlate tables with their foreign keys.
However, if no match is found in a case sensitive manner, there should be no harm to do a case *in*sensitive comparison as a fallback mechanism.

Also, at least a comment should be added, that explains why we might compare twice, so this does not get "optimized" away in the future (alternatively, the part could be extracted into its own meaningful named method).

dhirensham:
I do have a bit of a unorthodox environment, but I don't think that's causing the issue I experienced.

> What OS is your MySQL instance running under, which MySQL version are you using and what is the value of your lower_case_table_names status variable?

Server is MySQL 8.0.18 x64 running on macOS 10.15.2. lower_case_table_names value is 2.

> As far as I can see, this is only related to the MySQL Workbench specific option SQL Identifiers are Case Sensitive under Edit -> Preferences -> General Editors -> SQL Parsing in Code Editors. If the option is disabled, MySQL Workbench will use a lower case table name when adding foreign keys to a table.
> 
> I therefore assume, that you have/had the option disabled, correct?

I'm using Workbench 8.0.19 on Windows 10 1909 x64. But the `SQL Identifiers are Case Sensitive` option is ticked. Which points to

> This might be a bug in MySQL Workbench, because it should use the actual table name and not assume some casing when altering the table with added constraints. So this should also be reported on the MySQL Workbench bug tracker.

There does seem to be a bug in Workbench, where it's lower-casing the table names when generating scripts. I'll report it there as well. But that doesn't seem related to the actual problem that I encountered.

> As for the actual issue, we should discuss under wich circumstances this problem you try to fix arises.

When I switched from MySQL's EF Core library to Pomelo and re-ran the EF scaffold, I saw a number of messages in the console reporting that `Referenced table 'xxx' is not in dictionary.`. I also found that a number of navigation properties that I used to have were no longer being generated. It seemed related to the warning messages printed during the scaffold, so I dug through the source to see what was triggering it. 

Running the `GetConstraintsQuery` directly in Workbench gave me:
```
| CONSTRAINT_NAME          | TABLE_NAME         | REFERENCED_TABLE_NAME | PAIRED_COLUMNS | DELETE_RULE |
|--------------------------|--------------------|-----------------------|----------------|-------------|
| LoginPermissionMap_Login | LoginPermissionMap | login                 | LoginId|Id     | NO ACTION   |
```

So going back to the code, we're trying to find `FirstOrDefault(t => t.Name == referencedTableName)` which in this case is trying to compare `Login` with `login`, which fails.

lauxjpn:
I can reproduce this issue only with `SQL Identifiers are Case Sensitive` enabled.

Please test the necessary steps to reproduce the issue with `SQL Identifiers are Case Sensitive` enabled in a clean database with two new tables.

Then provide us with minimal step-by-step instructions, so we can replicate the same behavior on our end (or post us a quick video/animated gif).

dhirensham:
I had thought that it could have been Workbench on the Mac that had the `SQL Identifiers are Case Sensitive` option turned off, but it's on there too. I suspect what may have happened is that the table was created with the name `login` and later was renamed to `Login` for consistency, but the constraint was already created with the table name `login`.

Regardless, it seems that we're chasing the wrong thing here - you were able to reproduce the issue by unselecting the option in Workbench, which means that there are likely others who also have that option unselected who would have databases in this state. The connector from MySQL identified the constraints and generated the navigation properties as expected, it was only after switching to Pomelo that this became an issue. 

> I think this PR needs some alterations. Using a case sensitive comparison should still be the primary mechanism to correlate tables with their foreign keys.
However, if no match is found in a case sensitive manner, there should be no harm to do a case insensitive comparison as a fallback mechanism.
We should consider changing the insensitive comparison from FirstOrDefault() to Single(), to ensure that there is only one table that fits the name.

Happy to make this change and resubmit

lauxjpn:
I checked creating two tables with different casings on operating systems with and without case sensitive file name handling in conjunction with adding foreign keys that reference the parent table in a case sensitive and insensitive way.

The results are as follows:

On operating systems with sensitive file name handling, MySQL allows the creation of two tables with names that only differ in their casing (i.e. the tables `ParentTable` and `parenttable` are allowed to exists in parallel).
Adding a foreign key to a parent table with the not exact casing results in an error.

On operating systems with *in*sensitive file name handling, MySQL does not allow the creation of two tables with names that only differ in their casing (i.e. the tables `ParentTable` and `parenttable` are not allowed to exists in parallel, because they are represented by the same file).
Adding a foreign key to a parent table with the not exact casing does work.

So doing a case sensitive check and if unsuccessful an insensitive check as a fallback will work.
The alternative would be to check the `lower_case_table_names` status variable and then do the comparison accordingly (which would be more cumbersome than the former mechanism).
So I think we can keep the double check behavior.

The MySQL docs under [13.1.17.6 FOREIGN KEY Constraints](https://dev.mysql.com/doc/refman/5.6/en/create-table-foreign-keys.html) specify:

> The lower_case_table_names system variable setting is also taken into account.

The MySQL Workbench behavior is likely still a bug, though only on systems that use case sensitive file handling, because it could reference a table with the wrong casing (so this bug does actually not apply to our issue here).
 
> Regardless, it seems that we're chasing the wrong thing here

I think it is important that we know the scope of the problem, that we are implementing the fix for.

Were you able to reproduce the issue in the GUI with `SQL Identifiers are Case Sensitive` enabled, a new database and tables?

src/EFCore.MySql/Scaffolding/Internal/MySqlDatabaseModelFactory.cs
```diff
@@ -509,6 +509,12 @@ private void GetConstraints(DbConnection connection, IReadOnlyList<DatabaseTable
                         {
                             var referencedTableName = reader.GetString(2);
                             var referencedTable = tables.FirstOrDefault(t => t.Name == referencedTableName);
+                            if (referencedTable == null)
+                            {
+                                // if the casing of the constraint doesn't match the casing of the table name, try and find a matching table with a
+                                // different case, but ensure that there's only one possible match or we can't be certain which is the right table.
```
lauxjpn: Please change the comment to something that describes the issue we handle here. For example:

```c#
// On operation systems with insensitive file name handling, the saved reference table name might have a
// different casing than the actual table name. (#1017)
```

src/EFCore.MySql/Scaffolding/Internal/MySqlDatabaseModelFactory.cs
```diff
@@ -509,6 +509,12 @@ private void GetConstraints(DbConnection connection, IReadOnlyList<DatabaseTable
...
+                                referencedTable = tables.SingleOrDefault(t => string.Equals(t.Name, referencedTableName, StringComparison.OrdinalIgnoreCase));
```
lauxjpn: On the case *in*sensitive branch, we do not want any ambiguity between table names. The easiest way to archive this is to throw in case there are multiple tables that would satisfy the comparison (which should never happen).
So change `SingleOrDefault()` to `Single()`.

dhirensham:
> Were you able to reproduce the issue in the GUI with SQL Identifiers are Case Sensitive enabled, a new database and tables?

No, which again is confusing. But I was able to reproduce it as follows:

```
CREATE TABLE `test`.`Login` (
  `Id` INT NOT NULL,
  `LoginName` VARCHAR(45) NOT NULL,
  PRIMARY KEY (`Id`));

CREATE TABLE `test`.`LoginPermission` (
  `Id` INT NOT NULL,
  `LoginId` INT NOT NULL,
  PRIMARY KEY (`Id`),
  INDEX `LoginPermission_Login_idx` (`LoginId` ASC) VISIBLE,
  CONSTRAINT `LoginPermission_Login`
    FOREIGN KEY (`LoginId`)
    REFERENCES `test`.`login` (`Id`)
    ON DELETE NO ACTION
    ON UPDATE NO ACTION);
```

Note the `login` in the REFERENCES line, rather than `Login`. This results in:

```
C:\Projects\xxx\> dotnet ef dbcontext scaffold "server=mbp;port=3306;user=user;password=password;database=test" Pomelo.EntityFrameworkCore.MySql -c tempdb      
Build started...
Build succeeded.
Referenced table `login` is not in dictionary.
```

I was able to get Workbench to generate this by having 2 Create Table screens open at the same time, one creating table Login and the other creating table LoginPermission. Even if I then created Login, when editing the Foreign Keys for LoginPermission, Login doesn't appear in the Referenced Table list, so I can add it in by selecting Specify Manually... and entering `login`. So in this way, I was able to create table `Login` and table `LoginPermission` referencing `login`. 

A bit contrived I agree, but it seems plausible - create table1, forget to apply the changes, create table2 to reference table1, realise table1 hasn't been created yet, apply changes to table1, and then just specify table1 manually on table1, instead of re-entering all of the data you've already captured on table2.

src/EFCore.MySql/Scaffolding/Internal/MySqlDatabaseModelFactory.cs
dhirensham: SingleOrDefault has been changed to Single

lauxjpn:
Okay, that is in line with what I experienced as well. So manually (either by writing the name in the GUI or by using an `ALTER TABLE` statement directly) adding a reference with different casing (on a server with case insensitive file name handling) or disabling `SQL Identifiers are Case Sensitive` can lead to this result.

lauxjpn:
@dhirensham How is it going? We should be able to merge this after you added a test.

dhirensham:
Hi @lauxjpn 
Sorry, been tied up with other things. I've tried adding a test but am getting stuck. The best place for this one seems to be adding a test to `Scaffolding/MySqlDatabaseModelFactoryTest`. I've created a stub test to get started, but attempting to run it gives me the following:
```
Unable to cast object of type 'Microsoft.EntityFrameworkCore.DbContext' to type 'Microsoft.EntityFrameworkCore.TestUtilities.PoolableDbContext'.
```
when running this line:
```
            protected override string StoreName { get; } = nameof(MySqlDatabaseModelFactoryTest);
``` 
(technically, according to the stack trace it's crashing in the implicit MySqlDatabaseModelFixture constructor).

Any insight into what might be causing this?

For reference, this is what I've tried adding to the class:

```
        // Important: the inconsistent casing of the referenced table name is intentional. Do not change.
        [Fact]
        public void Ensure_constraints_scaffold_with_case_mismatch()
        {
            Test(
                @"
CREATE TABLE `PrincipalTable` (
  `Id` INT NOT NULL,
  PRIMARY KEY (`Id`));

CREATE TABLE `DependentTable` (
  `Id` INT NOT NULL,
  `ForeignKeyId` INT NOT NULL,
  PRIMARY KEY (`Id`),
  CONSTRAINT `ForeignKey_Id`
    FOREIGN KEY (`ForeignKeyId`)
    REFERENCES `principaltable` (`Id`)
);",
                Enumerable.Empty<string>(),
                Enumerable.Empty<string>(),
                dbModel =>
                {
                    Assert.True(true);
                },
                @"
DROP TABLE PrincipalTable;
DROP TABLE DependentTable;"
                );
        }
```

lauxjpn:
> Any insight into what might be causing this?

The following line needs to be changes:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/5cdca1859309fba5c167ab40fa027fa4c03bfa98/test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs#L820

It should really be:

```c#
public class MySqlDatabaseModelFixture : SharedStoreFixtureBase<PoolableDbContext>
```

dhirensham:
@lauxjpn Thanks, I've added the test.

test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs
```diff
@@ -731,6 +731,42 @@ public void Set_referential_action_for_foreign_key()
...
+                dbModel =>
+                {
+                    var principal = dbModel.Tables.FirstOrDefault(t => t.Name == "PrincipalTable");
+                    var dependent = dbModel.Tables.FirstOrDefault(t => t.Name == "DependentTable");
+
+                    Assert.NotNull(principal);
+                    Assert.NotNull(dependent);
+
+                    Assert.Contains(dependent.ForeignKeys, t => t.PrincipalTable.Name == principal.Name);
+                    Assert.True(true);
```
lauxjpn, line 763: Just remove this line and move the comment from outside the method inside the method, so it does not get lost later on. We can merge then.

dhirensham: Done

lauxjpn:
You can see, that you test currently fails on all 5 test environments in CI:
https://dev.azure.com/pomelo-efcore/Pomelo.EntityFrameworkCore.MySql/_build/results?buildId=580&view=results

Here are some of the error outputs:

**Linux with MySQL:**

```
[xUnit.net 00:01:31.12]     Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch [FAIL]
[xUnit.net 00:01:31.12]       MySql.Data.MySqlClient.MySqlException : Failed to open the referenced table 'principaltable'
[xUnit.net 00:01:31.12]       ---- MySql.Data.MySqlClient.MySqlException : Failed to open the referenced table 'principaltable'
[xUnit.net 00:01:31.12]       Stack Trace:
[xUnit.net 00:01:31.12]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs(116,0): at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet()
[xUnit.net 00:01:31.12]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs(391,0): at MySql.Data.MySqlClient.MySqlDataReader.CreateAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:01:31.13]         C:\projects\mysqlconnector\src\MySqlConnector\Core\CommandExecutor.cs(62,0): at MySqlConnector.Core.CommandExecutor.ExecuteReaderAsync(IReadOnlyList`1 commands, ICommandPayloadCreator payloadCreator, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:01:31.13]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlCommand.cs(226,0): at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:01:31.13]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlCommand.cs(74,0): at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(255,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.<>c.<ExecuteNonQuery>b__35_0(DbCommand command)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(234,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteCommand[T](DbConnection connection, Func`2 execute, String sql, Boolean useTransaction, Object[] parameters)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(212,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.Execute[T](DbConnection connection, Func`2 execute, String sql, Boolean useTransaction, Object[] parameters)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(255,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteNonQuery(DbConnection connection, String sql, Object[] parameters)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(252,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteNonQuery(String sql, Object[] parameters)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs(36,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Test(String createSql, IEnumerable`1 tables, IEnumerable`1 schemas, Action`1 asserter, String cleanupSql)
[xUnit.net 00:01:31.13]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs(768,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch()
[xUnit.net 00:01:31.13]         ----- Inner Stack Trace -----
[xUnit.net 00:01:31.13]         C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs(49,0): at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior)
```

**Linux with MariaDB:**

```
[xUnit.net 00:00:11.77]     Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch [FAIL]
[xUnit.net 00:00:11.77]       MySql.Data.MySqlClient.MySqlException : Can't create table `MySqlDatabaseModelFactoryTest`.`DependentTable` (errno: 150 "Foreign key constraint is incorrectly formed")
[xUnit.net 00:00:11.77]       ---- MySql.Data.MySqlClient.MySqlException : Can't create table `MySqlDatabaseModelFactoryTest`.`DependentTable` (errno: 150 "Foreign key constraint is incorrectly formed")
[xUnit.net 00:00:11.77]       Stack Trace:
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs(130,0): at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet()
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlDataReader.cs(391,0): at MySql.Data.MySqlClient.MySqlDataReader.CreateAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\Core\CommandExecutor.cs(62,0): at MySqlConnector.Core.CommandExecutor.ExecuteReaderAsync(IReadOnlyList`1 commands, ICommandPayloadCreator payloadCreator, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlCommand.cs(226,0): at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken)
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\MySql.Data.MySqlClient\MySqlCommand.cs(74,0): at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(255,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.<>c.<ExecuteNonQuery>b__35_0(DbCommand command)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(234,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteCommand[T](DbConnection connection, Func`2 execute, String sql, Boolean useTransaction, Object[] parameters)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(212,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.Execute[T](DbConnection connection, Func`2 execute, String sql, Boolean useTransaction, Object[] parameters)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(255,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteNonQuery(DbConnection connection, String sql, Object[] parameters)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/TestUtilities/MySqlTestStore.cs(252,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.TestUtilities.MySqlTestStore.ExecuteNonQuery(String sql, Object[] parameters)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs(36,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Test(String createSql, IEnumerable`1 tables, IEnumerable`1 schemas, Action`1 asserter, String cleanupSql)
[xUnit.net 00:00:11.78]         /home/vsts/work/1/s/test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs(768,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch()
[xUnit.net 00:00:11.78]         ----- Inner Stack Trace -----
[xUnit.net 00:00:11.78]         C:\projects\mysqlconnector\src\MySqlConnector\Core\ResultSet.cs(49,0): at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior)
```

**Windows with MySQL:**

```
[xUnit.net 00:01:06.23]     Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch [FAIL]
[xUnit.net 00:01:06.23]       Assert.NotNull() Failure
[xUnit.net 00:01:06.23]       Stack Trace:
[xUnit.net 00:01:06.24]         D:\a\1\s\test\EFCore.MySql.FunctionalTests\Scaffolding\MySqlDatabaseModelFactoryTest.cs(760,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.<>c.<Ensure_constraints_scaffold_with_case_mismatch>b__30_0(DatabaseModel dbModel)
[xUnit.net 00:01:06.24]         D:\a\1\s\test\EFCore.MySql.FunctionalTests\Scaffolding\MySqlDatabaseModelFactoryTest.cs(50,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Test(String createSql, IEnumerable`1 tables, IEnumerable`1 schemas, Action`1 asserter, String cleanupSql)
[xUnit.net 00:01:06.24]         D:\a\1\s\test\EFCore.MySql.FunctionalTests\Scaffolding\MySqlDatabaseModelFactoryTest.cs(768,0): at Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Scaffolding.MySqlDatabaseModelFactoryTest.Ensure_constraints_scaffold_with_case_mismatch()
```

So you can see, that the errors on Linux for MySQL and MariaDB are the same, just different messages, while the error on Windows is different.

### Linux

Our Linux systems use the following configuration (which is the default):

```
lower_case_file_system	OFF
lower_case_table_names	0
```

Here is the text from the [5.1.8 Server System Variables: lower_case_table_names](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) MySQL docs:

> If set to 0, table names are stored as specified and comparisons are case-sensitive. If set to 1, table names are stored in lowercase on disk and comparisons are not case sensitive. If set to 2, table names are stored as given but compared in lowercase. This option also applies to database names and table aliases. For additional details, see [Section 9.2.3, “Identifier Case Sensitivity”](https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html).
>
> On Windows the default value is 1. On macOS, the default value is 2. On Linux, a value of 2 is not supported; the server forces the value to 0 instead.

Since `lower_case_table_names=0` is used, `` the REFERENCES `principaltable` (`Id`) `` part will be correctly rejected by MySQL and MariaDB on Linux.

So you need to consider that in your test.

### Windows

Our Windows system used the following configuration on your test run:

```
lower_case_file_system	ON
lower_case_table_names	1
```

I did just change it to `lower_case_table_names=2` in #1062, so we run a wider variety of configurations in our CI test suite. However, `lower_case_table_names=1` is a valid (and the default) configuration for Windows and needs to work with your test as well.

Your test code currently checks the read table names in a case-sensitive way, which is not what you want here. You need to check them with `OrdinalIgnoreCase`. The case-sensitive `t.PrincipalTable.Name == principal.Name` check is fine though.

test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs
```diff
@@ -731,6 +731,41 @@ public void Set_referential_action_for_foreign_key()
...
+
+                    Assert.NotNull(principal);
+                    Assert.NotNull(dependent);
+
+                    Assert.Contains(dependent.ForeignKeys, t => t.PrincipalTable.Name == principal.Name);
+                },
+                @"
+DROP TABLE DependentTable;
+DROP TABLE PrincipalTable;"
+                );
```
lauxjpn, line 767: MySQL and MariaDB on Linux will correctly reject the `REFERENCES `principaltable` (`Id`)` (at least with `lower_case_table_names=0`). So this case needs to be handled here.

test/EFCore.MySql.FunctionalTests/Scaffolding/MySqlDatabaseModelFactoryTest.cs
```diff
@@ -731,6 +731,41 @@ public void Set_referential_action_for_foreign_key()
...
+  CONSTRAINT `ForeignKey_Id`
+    FOREIGN KEY (`ForeignKeyId`)
+    REFERENCES `principaltable` (`Id`)
+);",
+                Enumerable.Empty<string>(),
+                Enumerable.Empty<string>(),
+                dbModel =>
+                {
+                    var principal = dbModel.Tables.FirstOrDefault(t => t.Name == "PrincipalTable");
+                    var dependent = dbModel.Tables.FirstOrDefault(t => t.Name == "DependentTable");
```
lauxjpn, line 757: The table name checks need to be performed case-*in*sensitive (`OrdinalIgnoreCase`), because they would be e.g. saved as `principaltable` and `dependenttable` with `lower_case_table_names=1`.

lauxjpn:
Moved to #1144

PR #1056, opened 2020-04-01
Match Expression for MySQL 5.6+ and MariaDB 10.3+

neonVoice:
Good day to everyone! First time doing pull request.
I was trying to extend this library, because i fell in need of **Full Text Index** search using syntaxis from here: [https://mariadb.com/kb/en/match-against/](url).

I'm asking for help in order to understand how can i do all this stuff properly (contributing), and if not being able to push it to repository - what are possibilities for extension in that direction without tightly coupling my code to this library ?

P.S. Current implementation doesnt perform search on multiple columns inside MATCH operator, neither multiple tables, even though its not currently supported in MySQL as i know.


lauxjpn:
Nice one, thanks for contributing!

src/EFCore.MySql/Query/ExpressionTranslators/Internal/MySqlStringComparisonMethodTranslator.cs
```diff
@@ -229,6 +231,30 @@ public SqlExpression Translate(SqlExpression instance, MethodInfo method, IReadO
...
+        }
+
+        public SqlExpression MakeMatchWithQueryExpansionExpression(
+            [NotNull] SqlExpression target,
+            [NotNull] SqlExpression match,
+            [NotNull] SqlExpression against)
+        {
+            return _sqlExpressionFactory.MakeMatchWithQueryExpansion(match, against);
+        }
+
```
lauxjpn, line 257: Probably cleaner to have one method that takes a match search mode parameter.

src/EFCore.MySql/Query/Expressions/Internal/MySqlMatchExpression.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Pomelo.EntityFrameworkCore.MySql.Query.Expressions.Internal
+{
+    public enum MySqlMatchExpressionSearchMode
+    {
+        WithQueryExpansion,
+        InNaturalLanguageMode,
+        InNaturalLanguageModeWithQueryExpansion,
+        InBooleanMode
+    }
```
lauxjpn, line 17: Use this enum for the one public match method and rename it accordingly (e.g. to `MySqlMatchSearchMode`).

src/EFCore.MySql/Query/Internal/MySqlSqlExpressionFactory.cs
```diff
@@ -165,12 +220,12 @@ private SqlExpression ApplyTypeMappingOnRegexp(MySqlRegexpExpression regexpExpre
...
-                    resultType = left.Type;
-                    resultTypeMapping = inferredTypeMapping;
-                }
-                break;
+                    {
+                        inferredTypeMapping = typeMapping ?? ExpressionExtensions.InferTypeMapping(left, right);
+                        resultType = left.Type;
+                        resultTypeMapping = inferredTypeMapping;
+                    }
+                    break;
```
lauxjpn, line 228: Remove the change in formatting.

src/EFCore.MySql/Query/Internal/MySqlStringMethodTranslator.cs
```diff
@@ -50,6 +52,12 @@ public class MySqlStringMethodTranslator : IMethodCallTranslator
             = typeof(string).GetRuntimeMethod(nameof(string.Contains), new[] { typeof(string) });
         private static readonly MethodInfo _endsWithMethodInfo
             = typeof(string).GetRuntimeMethod(nameof(string.EndsWith), new[] { typeof(string) });
+        private static readonly MethodInfo _matchMethodInfo
+            = typeof(StringExt).GetRuntimeMethod(nameof(StringExt.Match), new[] {typeof(string), typeof(string)});
+        private static readonly MethodInfo _matchInBooleanModeMethodInfo
+            = typeof(StringExt).GetRuntimeMethod(nameof(StringExt.MatchInBooleanMode), new[] {typeof(string), typeof(string)});
+        private static readonly MethodInfo _matchWithQueryExpansionMethodInfo
+            = typeof(StringExt).GetRuntimeMethod(nameof(StringExt.MatchWithQueryExpansion), new[] {typeof(string), typeof(string)});
```
lauxjpn, line 60: I am not convinced that we should add a `Match()` string extension method. I think we should handle this similar to `LIKE`, that just has a `DbFunctions` extension method. So I think this should be removed.

If we decide to include this anyway, this should not be in `MySqlStringMethodTranslator`, but in a new method translator class.

src/EFCore.MySql/Query/Internal/MySqlStringMethodTranslator.cs
```diff
@@ -166,6 +174,24 @@ public virtual SqlExpression Translate(SqlExpression instance, MethodInfo method
                     .MakeEndsWithExpression(instance, arguments[0], _sqlExpressionFactory.Constant(StringComparison.CurrentCulture));
             }
 
+            if (_matchMethodInfo.Equals(method))
+            {
+                return new MySqlStringComparisonMethodTranslator(_sqlExpressionFactory)
+                    .MakeMatchExpression(instance, arguments[0], arguments[1]);
+            }
+
```
lauxjpn, line 182: I am not convinced that we should add a `Match()` string extension method. I think we should handle this similar to `LIKE`, that just has a `DbFunctions` extension method. So I think this should be removed.

If we decide to include this anyway, this should not be in `MySqlStringMethodTranslator`, but in a new method translator class.

src/EFCore.MySql/Query/Internal/MySqlStringMethodTranslator.cs
```diff
@@ -166,6 +174,24 @@ public virtual SqlExpression Translate(SqlExpression instance, MethodInfo method
...
+                return new MySqlStringComparisonMethodTranslator(_sqlExpressionFactory)
+                    .MakeMatchInBooleanModeExpression(instance, arguments[0], arguments[1]);
+            }
+
+            if (_matchWithQueryExpansionMethodInfo.Equals(method))
+            {
+                return new MySqlStringComparisonMethodTranslator(_sqlExpressionFactory)
+                    .MakeMatchWithQueryExpansionExpression(instance, arguments[0], arguments[1]);
+            }
+
```
lauxjpn, line 194: I am not convinced that we should add a `Match()` string extension method. I think we should handle this similar to `LIKE`, that just has a `DbFunctions` extension method. So I think this should be removed.

If we decide to include this anyway, this should not be in `MySqlStringMethodTranslator`, but in a new method translator class.

src/EFCore.MySql/Utilities/QueryableExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+            var command = sqlGenerator.GetCommand(selectExpression);
+
+            enumerator.Dispose();
+            return command.CommandText;
+        }
+
+        private static object Private(this object obj, string privateField) => obj?.GetType().GetField(privateField, BindingFlags.Instance | BindingFlags.NonPublic)?.GetValue(obj);
+        private static T Private<T>(this object obj, string privateField) => (T)obj?.GetType().GetField(privateField, BindingFlags.Instance | BindingFlags.NonPublic)?.GetValue(obj);
+    }
+}
```
lauxjpn, line 31: This seems not be used at all. Just remove it.

src/EFCore.MySql/Query/Internal/MySqlSqlExpressionFactory.cs
```diff
@@ -90,6 +91,45 @@ public MySqlSqlExpressionFactory(SqlExpressionFactoryDependencies dependencies)
...
+            MySqlMatchExpressionSearchMode searchMode = MySqlMatchExpressionSearchMode.InNaturalLanguageModeWithQueryExpansion)
+        {
+            return (MySqlMatchExpression)ApplyDefaultTypeMapping(
+                new MySqlMatchExpression(
+                    match,
+                    against,
+                    searchMode,
+                    null));
+        }
+
```
lauxjpn, line 132: Probably cleaner to have one method that takes a match search mode parameter.

src/EFCore.MySql/Utilities/StringExt.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            throw new NotImplementedException("Should not be called from client code.");
+        }
+
+        public static bool MatchWithQueryExpansion(this string match, string against)
+        {
+            throw new NotImplementedException("Should not be called from client code.");
+        }
+    }
+}
```
lauxjpn, line 24: I am not convinced that we should add a `Match()` string extension method. I think we should handle this similar to `LIKE`, that just has a `DbFunctions` extension method. So I think this should be removed.

If we decide to include this anyway, this should be renamed to something more expressive (e.g. `MySqlMatchStringExtensions`).

test/EFCore.MySql.FunctionalTests/Northwind.sql
```diff
@@ -270,6 +270,8 @@ CREATE TABLE `Products` (
 /* GO */
  CREATE  INDEX `SuppliersProducts` ON `Products`(`SupplierID`);
 /* GO */
+ CREATE FULLTEXT INDEX IF NOT EXISTS `FTI__Name` ON `Products`(`Name`)
+/* GO */
```
lauxjpn, line 274: It's probably not a good idea to extend the standard Northwind script. Instead, just create your own simple context (or copy an existing one) like many samples do.

test/EFCore.MySql.IntegrationTests/Tests/Models/ExpressionTest.cs
```diff
@@ -294,6 +301,33 @@ public async Task MySqlStartsWithOptimizedTranslator()
...
+
+        [Fact]
+        public async Task MySqlMatchWithQueryExpansionTranslator()
+        {
+            await Task.FromResult(1);
+            var result = _db.Products.Where(p => p.Name.MatchWithQueryExpansion("Product*"));
+
+            Assert.True(result.Any());
+        }
+
```
lauxjpn, line 330: Modify test names in accordance to the other tests in the test suite. Also assert for the generated SQL. This should be moved to the functional test project.

lauxjpn:
> I was trying to extend this library, because i fell in need of Full Text Index search using syntaxis from here: https://mariadb.com/kb/en/match-against/

This should also work against MySQL 5.6+, that uses the same syntax.

> I'm asking for help in order to understand how can i do all this stuff properly (contributing), and if not being able to push it to repository - what are possibilities for extension in that direction without tightly coupling my code to this library?

We are happy to include your PR in Pomelo! So no need to provide a plugin or something like that.

> P.S. Current implementation doesnt perform search on multiple columns inside MATCH operator, neither multiple tables, even though its not currently supported in MySQL as i know.

We could just add another overload to `DbFunctions.Match()`, that takes an array of `T`.

src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
```diff
@@ -146,6 +146,37 @@ public virtual Expression VisitMySqlRegexp(MySqlRegexpExpression mySqlRegexpExpr
...
+            Check.NotNull(mySqlMatchExpression, nameof(mySqlMatchExpression));
+
+            Sql.Append("MATCH ");
+            Sql.Append("(");
+            Visit(mySqlMatchExpression.Match);
+            Sql.Append(")");
+            Sql.Append(" AGAINST ");
+            Sql.Append($"(");
+            Visit(mySqlMatchExpression.Against);
+            //Sql.Append("`");
```
lauxjpn, line 160: Can be removed.

lauxjpn, line 160: Can be removed.

neonVoice:
Just to clarify my next actions: You made **Changes requested** on pull request, so i need to incorporat changes into my fork branch and make another pull request with fixed changes or i can somehow bind/link my next commit to **specific conversations**.

What do you usually have to do when you've fixed bugs, made adjustments, etc and pushed them to your fork branc, how to move from here ?
Should i make another pull request that should be linked to this pull request ?
While resolving conversations should i somehow include commit link to them ?

lauxjpn:
You can just add or change commits of your local repository. When you push them (or force-push them, which is fine on your fork) to your fork on GitHub (´neonVoice:master´), this PR (pull request) automatically gets updated.

test/EFCore.MySql.FunctionalTests/Northwind.sql
neonVoice: I removed changes from Northwind.sql, but not really sure where to create my own context, can you give me an example or point in right direction?

test/EFCore.MySql.IntegrationTests/Tests/Models/ExpressionTest.cs
neonVoice: Removed integration tests, will add functional tests when i'll understand how to create custom context with my index and test it. Also not sure about phrase **Also assert for the generated SQL**. Tests i have done were basic Read/Write tests to/from db, gonna look into it.

test/EFCore.MySql.FunctionalTests/Northwind.sql
lauxjpn: There are basically two approaches:

You can create your own `SharedStoreFixtureBase<PoolableDbContext>` derived fixture class, where you create your own data model. This gives you full control.
A random example would be the `OwnedQueryMySqlTest` class, that uses the `OwnedQueryMySqlFixture` class, that is derived from `RelationalOwnedQueryFixture`, which is derived from `OwnedQueryFixtureBase` and defines the data model (and finally derives from `SharedStoreFixtureBase<PoolableDbContext>`).
This can be done however, with just one fixture class directly deriving from `SharedStoreFixtureBase<PoolableDbContext>`.

The other way is to do something similar to `EscapesMySqlTest`, `EscapesMySqlFixture` and `EscapesMySqlFixtureBase`.
With `EscapesMySqlFixtureBase`, we just reuse some already defined `DbContext` (in this case `MusicStoreContext`). We then tailor the context to our needs (you can override `OnModelCreating()` to change the model) and change the `StoreName` property (so that a new database is created with our changed model and we don't change the behavior of existing tests in other classes).

test/EFCore.MySql.IntegrationTests/Tests/Models/ExpressionTest.cs
lauxjpn: > **Also assert for the generated SQL**. Tests i have done were basic Read/Write tests to/from db, gonna look into it.

I just mean to check the generated SQL against the expected one. E.g. all tests from the `SimpleQueryMySqlTest` class do this by calling the `AssertSql()` method at the end.

neonVoice:
From what i've understood, there is a **TestBase** class that can be modified by **Fixture** class + actual test class, that uses this **TestBase<TFixture>**.

Fixture allows me to change existing data model or create my own?

You said its better to go EF.DbFunctions.Like approach. So i found `DbFunctionsMySqlTest.MySql.cs` file with tests for **LIKE** operation, however, all of thouse tests use some kind of standard context and i plan on creating my own with FullTextIndex declared there. What name would be appropriate for test class with **MATCH** operations, which are DbFunctions too. Would something like `DbFunctionsExtensionsMySqlTest.MySql` ?

test/EFCore.MySql.IntegrationTests/Tests/Models/ExpressionTest.cs
neonVoice: You said its better to go **EF.DbFunctions.Like** approach. So i found `DbFunctionsMySqlTest.MySql.cs` file with tests for **LIKE** operation, however, all of thouse tests use some kind of standard context and i plan on creating my own with **FullTextIndex** declared there.
What name would be appropriate for test class with **MATCH** operations, which are **DbFunctions** too, something like `DbFunctionsExtensionsMySqlTest.MySql` is ok ?

test/EFCore.MySql.FunctionalTests/Query/MatchQueryMySqlTest.cs
```diff
@@ -0,0 +1,245 @@
...
+        public MatchQueryMySqlTest(MatchQueryMySqlFixture fixture)
+            : base(fixture)
+        {
+        }
+
+        [ConditionalFact]
+        public virtual void Match_In_Natural_Language_Mode()
+        {
+            using var context = CreateContext();
+            var count = context.Set<Herb>().Count(herb => EF.Functions.Match(herb.Name, "First", MySqlMatchSearchMode.InNaturalLanguageMode));
```
neonVoice, line 24: All of these tests show warning on MySqlMatchSearchMode, stating that it is internal API and it can change in the future. Any way to make it better ?

test/EFCore.MySql.IntegrationTests/Tests/Models/ExpressionTest.cs
lauxjpn: The name `MatchQueryMySqlTest` that you used is fine.

test/EFCore.MySql.FunctionalTests/Query/MatchQueryMySqlTest.cs
lauxjpn: You defined `MySqlMatchSearchMode` in the *internal* namespace `Pomelo.EntityFrameworkCore.MySql.Query.Expressions.Internal`:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/c89867b6fe9c514b49fe71dcf31ce904a63f847a/src/EFCore.MySql/Query/Expressions/Internal/MySqlMatchExpression.cs#L9-L17

Just define it in the same namespace as the `DbFunctions` extension methods (`Microsoft.EntityFrameworkCore`), since it is used as a parameter there.

You should also either move the `enum` to the `Extensions\MySqlDbFunctionsExtensions.cs` or create its own file under `Extensions\`.

src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
```diff
@@ -517,5 +515,36 @@ private static bool LikeCore<T>(T matchExpression, string pattern, string escape
...
+        public static bool Match<T>(
+            [CanBeNull] this DbFunctions _,
+            [CanBeNull] T matchExpression,
+            [CanBeNull] string pattern,
+            MySqlMatchSearchMode searchMode)
+        {
+            if (matchExpression is IConvertible convertible)
+            {
+                return EF.Functions.Match(convertible.ToString(CultureInfo.InvariantCulture), pattern, searchMode);
+            }
```
lauxjpn, line 545: This is a recursive call that would lead to a stack overflow. There is no .NET equivalent to a full-text search, so we should throw here.

There is a `FreeTextFunctionOnClient` resource string that is not used anymore, that you can reuse here.
You can change its value in `MySqlStrings.resx` from

```xml
<data name="FreeTextFunctionOnClient" xml:space="preserve">
  <value>The 'FreeText' method is not supported because the query has switched to client-evaluation. Inspect the log to determine which query expressions are triggering client-evaluation.</value>
</data>
```

to

```xml
<data name="FunctionOnClient" xml:space="preserve">
  <value>The '{methodName}' method is not supported because the query has switched to client-evaluation. Inspect the log to determine which query expressions are triggering client-evaluation.</value>
</data>
```

Then just run the T4 tool again on `MySqlStrings.Designer.tt` and rename all occurrences of `FreeTextFunctionOnClient` to `FunctionOnClient` in the solution.

Finally, change `MySqlDbFunctionsExtensions.Match<T>()` as follows:

```c#
public static bool Match<T>(
    [CanBeNull] this DbFunctions _,
    [CanBeNull] T matchExpression,
    [CanBeNull] string pattern,
    MySqlMatchSearchMode searchMode)
    => throw new InvalidOperationException(MySqlStrings.FunctionOnClient(nameof(Match)));
```

neonVoice: Just as i was confused with this `EF.Functions.Match`. Thanks for explanation, i was doing blind copy-paste, even though i didnt understood how would it work by calling itself, i thought that MethodCallTranslator will intercept it before going to client code somehow.

test/EFCore.MySql.FunctionalTests/Query/MatchQueryMySqlTest.cs
neonVoice: Haven't found enums in official dotnet/efcore repository to take example, so i moved it in separate file and changed namespace.

lauxjpn:
Looks good now! I am going to merge now.

I will push another PR in a bit, that will allow to not just use constant `MySqlMatchSearchMode` values, but also variables (parameters).

neonVoice:
It says - i can delete this fork now, but if i want to contribute more...

1. Do i need to make another fork for each issue ?
2. How to make sure that pull requests, associated with one issue, wont get associated with other issues ?
3. Basicaly how to work on 2 separate issues from one repository ? I need to make `feature` branch for each issue before start ?

PR #1061, opened 2020-04-10
Extend upon new MATCH...AGAINST support

lauxjpn:
This PR extends upon the work of @neonVoice's recent #1056 PR:

Add support for supplying the search mode as a parameter/variable instead of just as a constant value.
Simplify search mode names.
Remove redundant search mode (`IN NATURAL LANGUAGE MODE WITH QUERY EXPANSION` and `WITH QUERY EXPANSION` are the same mode).
Introduce explicit search mode values.

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlDbFunctionsEnums.cs
+++ b/src/EFCore.MySql/Extensions/MySqlDbFunctionsEnums.cs
@@ -2,9 +2,8 @@
 {
     public enum MySqlMatchSearchMode
     {
-        WithQueryExpansion,
-        InNaturalLanguageMode,
-        InNaturalLanguageModeWithQueryExpansion,
-        InBooleanMode
+        NaturalLanguage = 0,
+        NaturalLanguageWithQueryExpansion = 1,
+        Boolean = 2,
     }
 }
--- a/src/EFCore.MySql/Query/ExpressionTranslators/Internal/MySqlDbFunctionsExtensionsMethodTranslator.cs
+++ b/src/EFCore.MySql/Query/ExpressionTranslators/Internal/MySqlDbFunctionsExtensionsMethodTranslator.cs
@@ -112,40 +112,31 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionTranslators.Internal
             {
                 if (arguments[3] is SqlConstantExpression constant)
                 {
-                    return _sqlExpressionFactory.MakeMatch(arguments[1], arguments[2], (MySqlMatchSearchMode)constant.Value);
+                    return _sqlExpressionFactory.MakeMatch(
+                        arguments[1],
+                        arguments[2],
+                        (MySqlMatchSearchMode)constant.Value);
                 }
 
                 if (arguments[3] is SqlParameterExpression parameter)
                 {
-                    if (!_options.ServerVersion.SupportsMatchAgainstInsideCase)
-                    {
-                        // MariaDB is currently unable to perform a MATCH...AGAINST from inside of a CASE statement.
-                        // Use nested OR clauses instead:
-                        // <search_mode_1> = @p AND MATCH ... AGAINST ... OR
-                        // <search_mode_2> = @p AND MATCH ... AGAINST ... OR [...]
-                        var andClauses = Enum.GetValues(typeof(MySqlMatchSearchMode))
-                            .Cast<MySqlMatchSearchMode>()
-                            .OrderByDescending(m => m)
-                            .Select(m => _sqlExpressionFactory.AndAlso(_sqlExpressionFactory.Equal(parameter, _sqlExpressionFactory.Constant(m)), _sqlExpressionFactory.MakeMatch(arguments[1], arguments[2], m)))
-                            .ToArray();
-
-                        return andClauses
-                            .Skip(1)
-                            .Aggregate(
-                                andClauses.First(),
-                                (currentAnd, previousExpression) => _sqlExpressionFactory.OrElse(previousExpression, currentAnd));
-                    }
-
-                    var whenClauses = Enum.GetValues(typeof(MySqlMatchSearchMode))
+                    // Use nested OR clauses here, because MariaDB does not support MATCH...AGAINST from inside of
+                    // CASE statements and the nested OR clauses use the fulltext index, while using CASE does not:
+                    // <search_mode_1> = @p AND MATCH ... AGAINST ... OR
+                    // <search_mode_2> = @p AND MATCH ... AGAINST ... OR [...]
+                    var andClauses = Enum.GetValues(typeof(MySqlMatchSearchMode))
                         .Cast<MySqlMatchSearchMode>()
-                        .Where(m => m != default)
-                        .Select(m => new CaseWhenClause(_sqlExpressionFactory.Constant(m), _sqlExpressionFactory.MakeMatch(arguments[1], arguments[2], m)))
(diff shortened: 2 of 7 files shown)


neonVoice:
Is there any particular reason that azure tests batch is failing right after making first pull request, even though the code is right ?

lauxjpn:
In this case here, there actually seems to be a problem with the new test method and MariaDB. I have to recheck this locally.

Generally, the tests should run fine. In your case, there was some kind of communication error between the MySQL database (hosted in a docker container) and Pomelo. I haven't looked deeper into it. But if it happens again, we should investigate it further.

lauxjpn:
Okay, seems like MariaDB is unable to perform a match from inside of a `CASE`, though the case here would be evaluated just once before the query and only one of the branches would then be used to test all rows against. This is done by MySQL correctly.

lauxjpn:
I added special support for MariaDB. I will do some performance testing on the MySQL and the MariaDB query. If the MariaDB query is at least as fast as the MySQL query, we do not need to distinguish between database servers and can just use the MariaDB query for MySQL as well.

lauxjpn:
The MariaDB implementation is way faster on MySQL as well. It seems that the `CASE` statement does suppress the full text index usage.

neonVoice:
I tried executing this statement on my local project db and it yielded proper result

```SQL
SET @__searchMode_1 := '2';

SELECT COUNT(*)
FROM `herb` AS `h`
WHERE CASE @__searchMode_1
    WHEN 1 THEN MATCH (`h`.`Name`) AGAINST ('Арни*' WITH QUERY EXPANSION)
    WHEN 2 THEN MATCH (`h`.`Name`) AGAINST ('Арни*' IN BOOLEAN MODE)
    ELSE MATCH (`h`.`Name`) AGAINST ('Арни*')
END;
```

However, as you can see, im using cyrillic characters in against clause + structure of my projects db differs abit.

Here's DDL for my project DB:
```SQL
CREATE TABLE herb (
  Id int(11) NOT NULL AUTO_INCREMENT,
  Created datetime(6) NOT NULL,
  Modified datetime(6) NOT NULL,
  Name longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci DEFAULT NULL,
  PRIMARY KEY (Id)
)
ENGINE = INNODB,
AUTO_INCREMENT = 128,
AVG_ROW_LENGTH = 516,
CHARACTER SET utf8,
COLLATE utf8_general_ci;
```

And here's DDL for table, that was created in #1056
```SQL
CREATE TABLE herb (
  Id int(11) NOT NULL AUTO_INCREMENT,
  Name varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_general_ci DEFAULT NULL,
  PRIMARY KEY (Id)
)
ENGINE = INNODB,
AUTO_INCREMENT = 10,
AVG_ROW_LENGTH = 7281,
CHARACTER SET utf8mb4,
COLLATE utf8mb4_bin;
```

As you can see, there is a difference in definitions on column `Name`, they have different collations and types (`longtext` vs `varchar(255)`).
To summarize - `MATCH ... AGAINST` with `CASE` works in first example.

lauxjpn:
Whats your local database server (MySQL or MariaDB) and its version, where you ran your SQL on?
I also don't see any fulltext index defined.
What is the count you are getting?
After you defined the fulltext index, is it being used for the query?

neonVoice:
> Whats your local database server (MySQL or MariaDB) and its version (where you ran your SQL on)?

MariaDB 10.4.12
> I also don't see any fulltext index defined.

Just recreated db, applied full text index manualy through
`CREATE FULLTEXT INDEX FTI_herb_Name ON herb(Name);`
> What is the count you are getting?

well, for my local db proper count is 1 and select yields proper value. Haven't tried on data, that i included in test context (First, Second,...).
> After you defined the fulltext index for the `Name` column, is it being used for the query?

Ehm, yes ? Actualy when i tried to copy statement from AssertSql i got errors, because of missing trailing `;` after parameter declaration + missing `SET/SELECT/DECLARE` keywords for parameter.

lauxjpn:
> Ehm, yes ?

Does that mean yes, or are you unsure? What does the `EXPLAIN` statement return?

neonVoice:
> Does that mean yes, or are you unsure?
> What does the `EXPLAIN` statement return?
I'll answer later, in the morning.

neonVoice:
> What does the `EXPLAIN` statement return?

**Local Project DB**

Query:
```SQL
SET @__searchMode_1 := 2;

EXPLAIN EXTENDED
SELECT COUNT(*)
FROM `herb` AS `h`
WHERE CASE @__searchMode_1
    WHEN 1 THEN MATCH (`h`.`Name`) AGAINST ('Арни*' WITH QUERY EXPANSION)
    WHEN 2 THEN MATCH (`h`.`Name`) AGAINST ('Арни*' IN BOOLEAN MODE)
    ELSE MATCH (`h`.`Name`) AGAINST ('Арни*')
END;
```
Result:
| id | select_type | table | type | possible_keys | key    | key_len | ref    | rows | filtered | Extra       |
|----|-------------|-------|------|---------------|--------|---------|--------|------|----------|-------------|
| 1  | SIMPLE      | h     | ALL  | (null)        | (null) | (null)  | (null) | 127  | 100      | Using where |

**MatchQueryTest DB**

Query:
```SQL
SET @__searchMode_1 := 2;

EXPLAIN EXTENDED
SELECT COUNT(*)
FROM `herb` AS `h`
WHERE CASE @__searchMode_1
    WHEN 1 THEN MATCH (`h`.`Name`) AGAINST ('First*' WITH QUERY EXPANSION)
    WHEN 2 THEN MATCH (`h`.`Name`) AGAINST ('First*' IN BOOLEAN MODE)
    ELSE MATCH (`h`.`Name`) AGAINST ('First*')
END;
```

Result:
| id | select_type | table | type | possible_keys | key    | key_len | ref    | rows | filtered | Extra       |
|----|-------------|-------|------|---------------|--------|---------|--------|------|----------|-------------|
| 1  | SIMPLE      | h     | ALL  | (null)        | (null) | (null)  | (null) | 9    | 100      | Using where |

To summarize, results are identical.

lauxjpn:
Okay, you can see in your `EXPLAIN` output, that no index/key is being used in both queries. So this will be slow when run against a table with many records, like the one I created with the code example above that contains 100K records, because it uses a full table scan instead of the fulltext index. So using multiple `OR`/`AND` statements, as this PR does, is the correct approach.

Your results are interesting though. If you provide a full SQL script here (`CREATE DATABASE`, `CREATE TABLE` (with fulltext index), `INSERT INTO` (data), `SELECT` (the actual query)) I take a closer look at this.

neonVoice:
From what i know, there could be some kind of optimization during query execution, like in this situation its completely unnecesary to use index on such small tables (9 and 127 records), ill try to generate more data and test it.

lauxjpn:
> From what i know, there could be some kind of optimization during query execution, like in this situation its completely unnecesary to use index on such small tables (9 and 127 records), ill try to generate more data and test it.

No that is not the case here. The `possible_keys` field of your `EXPLAIN` result is `NULL`. So there are no keys available for MySQL to choose from.

If MySQL could have used a key, but chose not to do so, because it decided it would be faster to scan the table instead of using the index, the `possible_keys` field would not be empty, but the `key` field still would be.

neonVoice:
@lauxjpn Thanks for clarifying.

PR #1062, opened 2020-04-11
Improve database information output and configuration in CI

lauxjpn:
To support more advanced database test scenarios, we need to diversify our test environment a bit. E.g. the `lower_case_table_names` MySQL variable should be set to different values on Linux and Windows to test more scenarios for #1017.

Also updates all used versions to their latest releases.


lauxjpn:
Currently, the Chocolatey website (that we use to install MySQL for our Windows CI) is down (see their [status](https://status.chocolatey.org)). So Windows tests will fail until they are up and running again.

lauxjpn:
Chocolatey is back.

Issue #872, opened 2019-10-13
There is a problem with unique index


killerfrienddk:
```
Failed executing DbCommand (7ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
CREATE UNIQUE INDEX `RoleNameIndex` ON `AspNetRoles` (`NormalizedName`);

Failed executing DbCommand (8ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
CREATE INDEX `EmailIndex` ON `AspNetUsers` (`NormalizedEmail`);

Failed executing DbCommand (8ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
CREATE UNIQUE INDEX `RoleNameIndex` ON `AspNetRoles` (`NormalizedName`);

Failed executing DbCommand (6ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
CREATE UNIQUE INDEX `UserNameIndex` ON `AspNetUsers` (`NormalizedUserName`);
```

For all of them in the end:
`Specified key was too long; max key length is 767 bytes`

# Fix:
### From:
In 00000000000000_addCreateIdentity.Designer.cs file
```
b.Property<string>("NormalizedName")
.HasColumnType("varchar(256)")
.HasMaxLength(256);
```
In 00000000000000_addCreateIdentity.cs file
```
 NormalizedUserName = table.Column<string>(maxLength: 256, nullable: true)
```
### to:
```
b.Property<string>("NormalizedName")
.HasColumnType("varchar(255)")
.HasMaxLength(255);

 NormalizedUserName = table.Column<string>(maxLength: 255, nullable: true)
```

lauxjpn:
This is a limitation of MySQL and InnoDB, as can be read in the [official MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/create-index.html).

This is also explained for practical scenarios on [StackOverflow](https://stackoverflow.com/questions/1814532/1071-specified-key-was-too-long-max-key-length-is-767-bytes).

lauxjpn:
If anybody actually *needs* "Column Prefix Key Parts" implemented in Pomelo, feel free to open an issue.

Issue #380, opened 2017-09-07
Specified key was too long; max key length is 767 bytes


laosandegudai:
When i use it work well about default charse of mysql,but it dont work when i use it about utf8 of mysql.
MySql.Data.MySqlClient.MySqlException: Specified key was too long; max key length is 767 bytes
   at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet(ResultSet resultSet)
   at MySql.Data.MySqlClient.MySqlDataReader.<ReadFirstResultSetAsync>d__62.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.MySqlDataReader.<CreateAsync>d__61.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.CommandExecutors.TextCommandExecutor.<ExecuteReaderAsync>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.CommandExecutors.TextCommandExecutor.<ExecuteNonQueryAsync>d__1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.MySqlCommand.ExecuteNonQuery()
   at Microsoft.EntityFrameworkCore.Storage.Internal.MySqlRelationalCommand.<ExecuteAsync>d__3.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[...]

Issue #298, opened 2017-06-26
Specified key was too long; max key length is 767 bytes


yoyos:
### Steps to reproduce

- Create new ASPNETCORE 1.1 web project in visual studio with Basic User Authentication enabled.
- Add Pomelo.EntityFrameworkCore.MySQL nuget package and change your startup.cs to use MySQL it
- Update-Database.
- Error is shown.

### The issue

It seems, the error 'Specified key was too long; max key length is 767 bytes' is occurring when Migration Tool try to create those index.
When I comment those lines in the first migration, everything works.

            migrationBuilder.CreateIndex(
                name: "UserNameIndex",
                table: "AspNetUsers",
                column: "NormalizedUserName",
                unique: true);

            migrationBuilder.CreateIndex(
                name: "RoleNameIndex",
                table: "AspNetRoles",
                column: "NormalizedName",
                unique: true);
                
            migrationBuilder.CreateIndex(
                name: "IX_AspNetRoleClaims_RoleId",
                table: "AspNetRoleClaims",
                column: "RoleId");
                
            migrationBuilder.CreateIndex(
                name: "IX_AspNetUserClaims_UserId",
                table: "AspNetUserClaims",
                column: "UserId");

            migrationBuilder.CreateIndex(
                name: "IX_AspNetUserLogins_UserId",
                table: "AspNetUserLogins",
                column: "UserId");

### Full error

```
MySql.Data.MySqlClient.MySqlException: Specified key was too long; max key length is 767 bytes
   at MySql.Data.MySqlClient.MySqlDataReader.ActivateResultSet(ResultSet resultSet)
   at MySql.Data.MySqlClient.MySqlDataReader.<ReadFirstResultSetAsync>d__62.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at MySql.Data.MySqlClient.MySqlDataReader.<CreateAsync>d__61.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

Issue #259, opened 2017-04-15
MySQL Exception Handling


dansiegel:
# Issue

This was previously brought up in issue #200. When using AspNetCore.Identity, the entities start throwing an exception for the "Specified key was too long". The error thrown though is quite useless as it provides no context for debugging or fixing the entities.

# Expected Behavior

The exception thrown should include information at least for the Table and/or Property that caused the exception to be thrown. While I may have encountered this in a context that only contained the base Identity Context, it could very quickly become almost impossible to diagnose for a larger schema when you don't even know which model caused the exception to be thrown.

caleblloyd:
The solution here is to use Logging to see which statements are getting executed.

Change `Microsoft` log level to `Information` ([example](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/appsettings.json#L7))

If you don't already have this somewhere, here's some examples:

- [Read a JSON object into a configuration object](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/AppConfig.cs#L17-L31).
- [Add logging to MVC app in Startup.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/Startup.cs#L59)
- [Add logging to DBContext if using simple console app with no Startup.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/Pomelo.EntityFrameworkCore.MySql.PerfTests/AppDb.cs#L63)

This will print all SQL statements that are getting executed, and the last one printed will be the one where the exception occurred.

PR #1069, opened 2020-04-17
Implement support for index/key prefix lengths

### LINKED ISSUE #936: Add `HasKeyLength` method (for indices/keys)
### Steps to reproduce
It's migration of tables for Identity.

```
>              /*migrationBuilder.AlterColumn<string>(
                name: "Id", <-- It's PrimaryKey!
                table: "AspNetUsers",
                nullable: false,
                oldClrType: typeof(string),
                oldType: "varchar(255)");*/
> 
>             /*migrationBuilder.AlterColumn<string>(
>                 name: "LoginProvider", <-- It's Index!
>                 table: "AspNetUserTokens",
>                 nullable: false,
>                 oldClrType: typeof(string),
>                 oldType: "varchar(255)");*/
> 
>             /*migrationBuilder.AlterColumn<string>(
>                 name: "UserId", <-- It's ForeignKey!
>                 table: "AspNetUserTokens",
>                 nullable: false,
>                 oldClrType: typeof(string),
>                 oldType: "varchar(255)");*/
```

### The issue
**LongText can't be PrimaryKey, ForeignKey and Index!**

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
Exception message:
Stack trace:
MySql.Data.MySqlClient.MySqlException (0x80004005): Specified key was too long; max key length is 1000 bytes
 ---> MySql.Data.MySqlClient.MySqlException (0x80004005): Specified key was too long; max key length is 1000 bytes
   at MySqlConnector.Core.ResultSet.ReadResultSetHeaderAsync(IOBehavior ioBehavior)

### Further technical details

MySQL version: 8.0
Operating system: Windows
Pomelo.EntityFrameworkCore.MySql version: 3.0.0-rc3.final
Microsoft.AspNetCore.App version: 3.0

### PR DESCRIPTION
**Pull request by lauxjpn**
The following code can be used to specify prefix lengths for indices:

```c#
// One column:
entity.HasIndex(e => e.SingleProperty)
    .HasPrefixLength(48);

// Multiple columns:
entity.HasIndex(e => new { e.FirstProperty, e.SecondProperty })
    .HasPrefixLength(0, 32);
```

The code to specify prefix lengths for keys is similar:

```c#
// One column:
entity.HasKey(e => e.SingleProperty)
    .HasPrefixLength(48);

// Multiple columns:
entity.HasKey(e => new { e.FirstProperty, e.SecondProperty })
    .HasPrefixLength(0, 32);
```

If you want to specify the usage of the whole length of a column as part of an index/key, specify the length as `0`, as can be seen in the `Multiple columns` examples.

~~I am not sure, if we should use the singular `HasPrefixLength`, instead of the plural.~~

Closes #936 

Also addresses #872, #380, #298, #259, #136, #99, #81 and #7

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlIndexBuilderExtensions.cs
+++ b/src/EFCore.MySql/Extensions/MySqlIndexBuilderExtensions.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Pomelo Foundation. All rights reserved.
 // Licensed under the MIT. See LICENSE in the project root for license information.
 
+using System;
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 using Microsoft.EntityFrameworkCore.Utilities;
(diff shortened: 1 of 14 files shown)


mguinness:
> I am not sure, if we should use the singular `HasPrefixLength`, instead of the plural.

I would go for singular as it's consistent with all the other method names.

olicooper:
I opened an issue with EFCore team but they pointed me to you guys.. [original issue](https://github.com/dotnet/efcore/issues/20751#issuecomment-619364616)

Could you extend this to support `index` prefix lengths too? From the example above this is specific to keys only?

olicooper:
Currently when creating an index, the migration builder converts string properties to `varchar(255)` when adding an index using the FluentAPI. Will we be able to use this new functionality without the builder converting `LONGTEXT` columns to `varchar(255)`?

lauxjpn:
@olicooper Good point! We should remove the current behavior to limit the key, now that the prefix can be set. This was always more of a hack anyway. However, since this would be a breaking change, we will remove this only for the .NET/EF Core 5 release.

Issue #376, opened 2017-09-02
An item with the same key has already been added. Key: Microsoft.EntityFrameworkCore.Infrastructure.Internal.MySqlOptionsExtension


n-gao:
### The issue
I am not able to create a migration to update my mysql database. It throws an exception.

```
Exception message:
Stack trace: 

info: Microsoft.AspNetCore.DataProtection.KeyManagement.XmlKeyManager[0]
      User profile is available. Using 'C:\Users\nicho\AppData\Local\ASP.NET\DataProtection-Keys' as key repository and Windows DPAPI to encrypt keys at rest.
System.ArgumentException: An item with the same key has already been added. Key: Microsoft.EntityFrameworkCore.Infrastructure.Internal.MySqlOptionsExtension
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector)
   at Microsoft.EntityFrameworkCore.DbContextOptions`1.WithExtension[TExtension](TExtension extension)
   at Microsoft.EntityFrameworkCore.DbContextOptionsBuilder.Microsoft.EntityFrameworkCore.Infrastructure.IDbContextOptionsBuilderInfrastructure.AddOrUpdateExtension[TExtension](TExtension extension
)
   at Microsoft.EntityFrameworkCore.MySqlDbContextOptionsExtensions.ConfigureWarnings(DbContextOptionsBuilder optionsBuilder)
   at Microsoft.EntityFrameworkCore.MySqlDbContextOptionsExtensions.UseMySql(DbContextOptionsBuilder optionsBuilder, String connectionString, Action`1 mySqlOptionsAction)
[...]

Issue #356, opened 2017-08-16
An item with the same key has already been added. Key: p2068


falps:
你好，使用Pomelo.EntityFrameworkCore.MySql（2.0.0-rtm-10056），在执行context.SaveChanges();时报错，错误内容：An item with the same key has already been added. Key: p2068，
StackTrace如下
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.CreateStoreCommand()
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.Update.Internal.MySqlBatchExecutor.Execute(IEnumerable`1 commandBatches, IRelationalConnection connection)
   at Microsoft.EntityFrameworkCore.Storage.RelationalDatabase.SaveChanges(IReadOnlyList`1 entries)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(IReadOnlyList`1 entriesToSave)
   at Microsoft.EntityFrameworkCore.ChangeTracking.Internal.StateManager.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges(Boolean acceptAllChangesOnSuccess)
   at Microsoft.EntityFrameworkCore.DbContext.SaveChanges()
同样的代码，使用版本：Pomelo.EntityFrameworkCore.MySql（1.1.2）没有问题，请帮忙看看，谢谢

Kagamine:
Please provide a repro for your issue, and use English to describe your issue.

munir-jb:
hi guys, I found these same bug because of the inability of .UseMySQL method to automatically checking wether optionsBuilder.Isconfigured in an EF class, so I had to do this :

`      try {
        if (!optionsBuilder.IsConfigured) {
          optionsBuilder.UseMySql("connection string");
          CustomizeConfiguration(ref optionsBuilder);
          base.OnConfiguring(optionsBuilder);
        }
      }
      catch (Exception e) {
[...]

PR #1075, opened 2020-04-27
An item with the same key has already been added. Key: Microsoft.Enti…

Prince269090:
Fixed issue #376
Fixed issue #356

I have fixed the issue as it was missing the casting of 'MySqlOptionsExtension' class. 

1. So first time if object dosen't exists then GetOrCreateExtension returns 'MySqlOptionsExtension' class. 

2. But second time if it exists then it returns the base class of 'MySqlOptionsExtension' which is RelationalOptionsExtension.

That's why update extension was not happening. 

Check the same code in entity framwork core for sql server [here](https://github.com/dotnet/efcore/blob/master/src/EFCore.SqlServer/Extensions/SqlServerDbContextOptionsBuilderExtensions.cs).

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlDbContextOptionsExtensions.cs
+++ b/src/EFCore.MySql/Extensions/MySqlDbContextOptionsExtensions.cs
@@ -31,7 +31,7 @@ namespace Microsoft.EntityFrameworkCore
             };
 
             connectionString = csb.ConnectionString;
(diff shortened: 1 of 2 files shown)


AndriySvyryd:
@Prince269090 Please add a test that fails without this change

lauxjpn:
@Prince269090 Nice find!

AndriySvyryd (approved):

PR #1080, opened 2020-05-01
Don't cast if not necessary

### LINKED ISSUE #1072: Redundant CAST() from and to DATETIME(6) might be generated, when System.DateTime and System.DateTimeOffset are used in conjunction
### Steps to reproduce
Create a table with datetime(6) column in MySQL 5.7.
Add index to this column.
Fill some data.

Then from EF Core:
await _dbContext.AuditTable.Where(n => n.CreatedAt >= startDate && n.CreatedAt < endDate).ToListAsync();

### The issue
It created a select like this before 3.x version:
select column1, column2 from AuditTable where CreatedAt >= @expr1 and CreatedAt < @expr2

Now, in 3.x the select looks like this:

select column1, column2 from AuditTable where CAST(CreatedAt as datetime(6)) >= @expr1 and CAST(CreatedAt as datetime(6)) < @expr2

The problem is, that in this case MySql cannot use the index, so if the AudiTable is about 22GB, then all query will timout.

### Further technical details

MySQL version: MySQL 5.7 (And in AWS Aurora last version)
Operating system: Windows 10 (and .core docker 3.1.2 in AWS)
Pomelo.EntityFrameworkCore.MySql version: tested with 3.0.0 and above (including nightly build from yesterday)
Microsoft.AspNetCore.App version: 3.1.2

lauxjpn:
Don't cast if not necessary. This might be the case, if `System.DateTime` and `System.DateTimeOffset` are used together, which both can map to the same store type (e.g. `datetime(6)`).

Implements a fix based on the one originally proposed by @Meister1977.

Fixes #1072

### DIFF
--- a/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
+++ b/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
@@ -236,27 +236,37 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
                 castMapping = "decimal(65,30)";
             }
 
-            if (useDecimalToDoubleWorkaround)
+            // There needs to be no CAST() applied between the exact same store type. This could happen, if
+            // `System.DateTime` and `System.DateTimeOffset` are used in conjunction, because both use different type
+            // mappings, but map to the same store type (e.g. `datetime(6)`).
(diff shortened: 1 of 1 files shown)


Meister1977:
Thanks. Do you plan to release a 3.1.2 version? Because this issue is blocker in our systems.

lauxjpn:
The next planned release is `3.2.0` later this month. You can also use our latest [nightly build](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#nightly-builds), or merge this particular PR locally and compile your own version.

Issue #1088, opened 2020-05-25
EF Core 5 support


timmi-on-rails:
.NET 5 is coming and the release is planned for November 2020.
At the same time EF Core 5 will be released.
Today I can already try the preview builds of .NET 5 and EF Core 5 via NuGet.
But I am unable to use the Pomelo.EntityFrameworkCore.MySql adapter, since there are breaking changes in EF Core 5.
I couldn't find a build of the adapter targeting EF Core 5 Preview.

Do you have plans for that and what is the progress?
I saw there is a 5.0-wip branch.

lauxjpn:
An EF Core 5 preview compatible Pomelo release is planned for the second half of June.

timmi-on-rails:
Any news? On what branch are you developing the .net 5.0 upgrade?
Maybe I am able to contribute, to speed things up.
5.0-wip branch hasn't been touched since February.

mguinness:
The current effort is focused on adding spatial and JSON support to 3.x and then afterwards upgrading will be looked at.  Based on previous version upgrades it's a significant amount of work which will take several months based on the limited developer resources.

ansarizafar:
>An EF Core 5 preview compatible Pomelo release is planned for the second half of June.

@lauxjpn  June is almost over. Can we expect An EF Core 5 preview compatible release this month.

timmi-on-rails:
Based on version 3.1.1 I got an upgrade to EF Core 5 preview "working".
I have inserted some records and ran a select statement successfully.

I still have to run the tests and I need help at 2-3 places.
Just wanted to inform you, that it is possible to get a "compiling" version in couple hours.
Probably faster, when knowing EF Core + database provider api.

EF Core 5 does not require net5.0, only netstandard2.1.
Ofcourse this can still change, because it is preview state, still.

lauxjpn:
We will publish an official preview after the JSON support release, of which I have a prototype running at the moment, that still needs a bit of polishing.
So expect an EF Core 5.0 preview compatible release until about 2020-07-15.

timmi-on-rails:
@lauxjpn I am currently working on a project, which is still in an early phase and the first release is planned for end of 2020, so I wanted to update the project dependencies, e.g. efcore to the preview of 5.0, because this will be the up-to-date version, when the project is released.
The reason is to test and find incompatibilities early as possible.
I haven't checked if any new efcore 5 features are useful for the project.
[...]

PR #1103, opened 2020-06-14
Update to MySqlConnector 1.0

bgrainger:
MySqlConnector 1.0 changes the primary namespace to `MySqlConnector`; this updates all relevant code.

This starts the discussion around updating MySqlConnector to 1.0, but doesn't need to be merged until it's out of beta.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -29,7 +29,7 @@
     <MicrosoftNETCoreAppInternalPackageVersion>3.1.3-servicing.20128.1</MicrosoftNETCoreAppInternalPackageVersion>
     <!-- EFCore.MySql Dependencies -->
     <MicrosoftEntityFrameworkCoreRelationalVersion>3.1.3</MicrosoftEntityFrameworkCoreRelationalVersion>
-    <MySqlConnectorVersion>0.69.2</MySqlConnectorVersion>
+    <MySqlConnectorVersion>1.0.0</MySqlConnectorVersion>
     <PomeloJsonObjectVersion>2.2.1</PomeloJsonObjectVersion>
     <MicrosoftExtensionsCachingMemoryVersion>3.1.3</MicrosoftExtensionsCachingMemoryVersion>
     <MicrosoftExtensionsDependencyInjection>3.1.3</MicrosoftExtensionsDependencyInjection>
--- a/src/EFCore.MySql.NTS/Storage/Internal/MySqlGeometryTypeMapping.cs
+++ b/src/EFCore.MySql.NTS/Storage/Internal/MySqlGeometryTypeMapping.cs
@@ -6,10 +6,9 @@ using System.Data.Common;
 using System.Reflection;
 using System.Text;
 using JetBrains.Annotations;
-using MySql.Data.MySqlClient; // Note: Hard reference to MySqlClient here.
 using Microsoft.EntityFrameworkCore.Storage;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
-using MySql.Data.Types;
+using MySqlConnector;
 using NetTopologySuite;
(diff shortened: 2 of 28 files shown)


bgrainger:
Upstream issue: https://github.com/mysql-net/MySqlConnector/issues/824.

lauxjpn:
I think it is a good move, as long as the namespaces can be corrected in the user code by using find & replace for a couple of releases.

Since this will be a breaking change, we should make this PR part of the `5.0.0` branch, and let `3.x.x` continue using the last available `MySql.Data` namespace release.

For backward compatibility reasons, I think it would be best to have a maintenance branch in MySqlConnector for `0.x.x`, in case some critical bug fixes need to be shipped for older versions (e.g. Pomelo 3.2.x will stay aligned with .NET Core 3.1, which is an LTS release and will be around for some time).

bgrainger:
> For backward compatibility reasons, I think it would be best to have a maintenance branch in MySqlConnector for 0.x.x

I haven't created a maintenance branch yet, but I plan to do so when the need arises.

bgrainger:
PR updated to MySqlConnector 1.0.0 (RTM).

MySqlConnector has a [`v0.x` branch](https://github.com/mysql-net/MySqlConnector/tree/v0.x) for maintenance.

trenttobler:
It looks like the current NUGET package for the Pomelo EF library uses a dependency:
MySqlConnector (>= 0.61.0)

This caused as a bit of an issue to track down the reason after we upgraded to MySqlConnector 1.0.0, as to why we started seeing runtime issues (instead of compile time version warnings / checks).

Asking to see if that version check can include <1.0.0, in addition to releasing this pull, to have least have a chance of finding the incompatibility a little easier?  Ultimately, we were able to work around this by downgrading to latest 0.x version of MySqlConnector.

bgrainger:
@lauxjpn the current Pomelo version should probably be republished with `<PackageReference Include="MySqlConnector" Version="[0.61.0, 1.0.0)" />` (or a higher starting version) to prevent undesirable upgrades.

bgrainger:
I've opened a PR to try to fix this problem: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1126

lauxjpn:
> @lauxjpn the current Pomelo version should probably be republished with <PackageReference Include="MySqlConnector" Version="[0.61.0, 1.0.0)" /> (or a higher starting version) to prevent undesirable upgrades.

@bgrainger Can a package be republished with the same version number? If so, what are the implications for users who already downloaded the previously published package? Or should we make it a new patch release (`3.1.2`)?

bgrainger:
@lauxjpn Sorry, I was imprecise. A package version can't be reused/updated. A _new_ version (with a patch version bump) should be published.

KevM:
Just checking in on this. I have a project which was using Pomelo and a separate lib had a package reference to MySqlConnector >= 1.0. I get runtime exceptions in a aspnetcore startup `.UseMySql(...)` because it could not find the right `MySqlConnectionStringBuilder` in the old pre 1.0 namespace. (Not sure why that is getting loaded via reflection). I am guessing this update would fix that problem.

lauxjpn:
@KevM Pomelo 3.x will not use MySqlConnector >= 1.0.0, because it would break SemVer2 compliance. All recent releases of Pomelo explicitly restrict the version of MySqlConnector to < 1.0.0, so you get a compiler error instead of a runtime error.

However, Pomelo 5.0 will use MySqlConnector >= 1.0.0. It is likely that the first alpha will make it into our nightly build feed today.

This is tracked specifically by #1161 and more generally by #1088.

lauxjpn:
This has already been done as part of #1182.

PR #1118, opened 2020-07-08
Add support for AUTO_INCREMENT in conjunction with value converters

### LINKED ISSUE #1117: AUTOINCREMENT not generated on primary key with ValueConverter
### Steps to reproduce
Create a console NET Core 3.1 project.
**TestEntityCustom.cs**
```cs
namespace TestEFCore
{
    public readonly struct TestId
    {
        private readonly uint id;

        private TestId(uint id)
        {
            this.id = id;
        }

        public uint ToId() => this.id;
        public static TestId From(uint id) => new TestId(id);
    }

    public class TestEntityCustom
    {
        public TestId Id { get; set; }
    }
}
```

**TestDbContext.cs**
```cs

using System;

using Pomelo.EntityFrameworkCore.MySql.Infrastructure;

namespace TestEFCore
{
    public class TestDbContext : DbContext
    {
        public DbSet<TestEntityCustom> TestCustomEntities { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            optionsBuilder
                // replace with your connection string
                .UseMySql(
                    "server=localhost;port=33060;user=xxxxx;password=yyyy;database=dbtest;charset=latin1",
                    mySqlOptions => mySqlOptions
                    // replace with your Server Version and Type
                        .ServerVersion(new Version(8, 0, 18), ServerType.MySql));
        }

        protected override void OnModelCreating(ModelBuilder modelBuilder)
        {
            modelBuilder.Entity<TestEntityCustom>()
                .Property(e => e.Id)
                .HasConversion(e => e.ToId(), e => TestId.From(e));
        }
    }
}
```

**program.cs**
```cs
using System;

using Microsoft.Extensions.DependencyInjection;

using Pomelo.EntityFrameworkCore.MySql.Infrastructure;

namespace TestEFCore
{
    class Program
    {
        static void Main()
        {
        }
    }
}
```

### The issue
The AUTOINCREMENT statement is not generated in the SQL script when running `dotnet ef migrations script`. 
```sql
CREATE TABLE `TestCustomEntities` (
    `Id` int unsigned NOT NULL,
    CONSTRAINT `PK_TestCustomEntities` PRIMARY KEY (`Id`)
);
```

I've added `HasKey` et `ValueGeneratedOnAdd` when creating the model with no result.
If i replace `TestId` by `uint`, it work flawlessly.

Maybe related to #720 

### Further technical details

MySQL version: 5.7.28
Operating system: ubuntu 18.04 (server) / Win10 (client)
Pomelo.EntityFrameworkCore.MySql version: 3.1.1
Microsoft.AspNetCore.App version: 3.1.5

lauxjpn:
Fixes #1117

Also an issue in other providers (Npgsql, SQL Server).
/cc @roji


src/EFCore.MySql/Properties/MySqlStrings.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="IdentityBadType" xml:space="preserve">
-    <value>Identity value generation cannot be used for the property '{property}' on entity type '{entityType}' because the property type is '{propertyType}'. Identity value generation can only be used with signed integer, DateTime, and DateTimeOffset properties.</value>
+    <value>Identity value generation cannot be used for the property '{property}' on entity type '{entityType}' because the property type is '{propertyType}'. Identity value generation can only be used with integer, DateTime, and DateTimeOffset properties.</value>
```
roji, line 121: Am curious, does MySQL actually have unsigned integer types, and do those support AUTO_INCREMENT?

roji, line 121: BTW I think the resx has been changed here (with some unrelated whitespace changes), but not the generated cs from it (the new message won't actually be applied until that is done).

src/EFCore.MySql/Extensions/MySqlPropertyExtensions.cs
```diff
@@ -102,13 +102,12 @@ private static void CheckValueGenerationStrategy(IProperty property, MySqlValueG
...
-            return (type.IsInteger()
-                        || type == typeof(decimal)
-                        || type == typeof(DateTime)
-                        || type == typeof(DateTimeOffset))
-                   && !HasConverter(property);
+            var valueConverter = GetConverter(property);
+            var type = (valueConverter?.ProviderClrType ?? property.ClrType).UnwrapNullableType();
+            return type.IsInteger()
+                   || type == typeof(decimal)
+                   || type == typeof(DateTime)
```
roji, line 109: Wow, AUTO_INCREMENT on date/time types?

roji:
LGTM, @ajcvickers any comments on this?

src/EFCore.MySql/Properties/MySqlStrings.resx
lauxjpn: Yes, it is even somehow encouraged (from [3.6.9 Using AUTO_INCREMENT](https://dev.mysql.com/doc/refman/8.0/en/example-auto-increment.html)):

> Use the smallest integer data type for the AUTO_INCREMENT column that is large enough to hold the maximum sequence value you will need. When the column reaches the upper limit of the data type, the next attempt to generate a sequence number fails. **Use the UNSIGNED attribute if possible to allow a greater range.**

src/EFCore.MySql/Extensions/MySqlPropertyExtensions.cs
lauxjpn: Not really, we are using (or misusing) it here as a way to allow users to define a `CURRENT_TIMESTAMP` as the `DEFAULT` for a column. This can currently be done in multiple ways, one of them is using `ValueGeneratedOnAdd()` on a `DateTime` column.

src/EFCore.MySql/Properties/MySqlStrings.resx
roji: First time I hear of first-class unsigned type support in a relational database... Thanks for the info!

lauxjpn: Hmm, its true that I did not regenerate the CS file (doesn't work from Rider, so I would need to load the project in VS to do so). But shouldn't the message be applied anyhow? After all, it is part of the resources, from which it should be retrieved at runtime. As far as I can see, only the description (comment) in the autogen CS file would be updated, when I let T4 run in VS in this case here (not that that is not a good idea).

src/EFCore.MySql/Extensions/MySqlPropertyExtensions.cs
lauxjpn: This will make more sense here, when we update our value generation related code for 5.0 and rename `IsCompatibleIdentityColumn` to `IsCompatibleWithValueGeneration`, as Npgsql and MSSQL already did for 3.0.

src/EFCore.MySql/Properties/MySqlStrings.resx
roji: You may be right - I usually manually update both the T4 and the auto-generated, have never looked too closely at this...

roji:
Looks like this needs to be fixed in SQL Server: https://github.com/dotnet/efcore/blob/7c572c5f570aff771586bf08c57aa73c424d4730/src/EFCore.SqlServer/Extensions/SqlServerPropertyExtensions.cs#L433

@lauxjpn feel like submitting a PR for this to EF Core or is one provider enough? :)

roji:
For EFCore.PG: https://github.com/npgsql/efcore.pg/issues/1439

lauxjpn:
> Looks like this needs to be fixed in SQL Server: https://github.com/dotnet/efcore/blob/7c572c5f570aff771586bf08c57aa73c424d4730/src/EFCore.SqlServer/Extensions/SqlServerPropertyExtensions.cs#L433
> 
> @lauxjpn feel like submitting a PR for this to EF Core or is one provider enough? :)

@roji Sure, I'll get on it.

roji:
@lauxjpn we didn't end up addressing this in 5.0, right?

lauxjpn:
@roji That is correct. Are the doors closed on the 5.0 release?

roji:
Yeah, they are - only critical bug fixes can make it in at this point...

This can still be fixed in your provider (and mine) if we think it's needed by people, or we can wait and do it across the board for 6.0.

PR #1119, opened 2020-07-09
Implement HEX() and UNHEX() support as EF.Functions

### LINKED ISSUE #1111: How to use HEX from EFCore?
### Steps to reproduce
I can write an SQL query like this to search in a byte array as if it was a string.
```
select HEX(SystemId) from Table where HEX(SystemId) like '%8b%'
```
Here `SystemId` is a non-nullable `longblob` (`byte[]` in *C#*).

### The issue
Is there a way to generate this query using EFCore?

### Further technical details

MySQL version: mysql  Ver 15.1 Distrib 10.4.12-MariaDB
Operating system: Linux
Pomelo.EntityFrameworkCore.MySql version: 3.1.1
Microsoft.AspNetCore.App version: 3.1.3

lauxjpn:
Implements client-side and server-side support for `HEX()` and `UNHEX()` as `EF.Functions.Hex()` and `EF.Functions.Unhex()`.

Fixes #1111

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
+++ b/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Globalization;
+using System.Text;
 using JetBrains.Annotations;
 using Pomelo.EntityFrameworkCore.MySql.Internal;
 
+// ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


mc0re:
That's great!
From the code I can't deduce whether HEX will work on a byte array. This is the primary goal of my original question, as comparing strings is not a problem.

lauxjpn:
According to the [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_hex), `HEX()` works on strings and numbers.

mc0re:
As far as I understand, `BINARY` type is a kind of string: https://docs.oracle.com/cd/E17952_01/mysql-5.1-en/binary-varbinary.html
`UNHEX` explicitly returns a *binary string*: https://www.w3resource.com/mysql/string-functions/mysql-unhex-function.php, so one can assume `HEX` accepts a *binary string* :-)

But in software world `BINARY` is a representation of `byte[]`, not `string`. Therefore my question.

lauxjpn:
Feel free to try it out and report back to us with your findings.
At least the following should work:

```sql
select HEX(CAST(`BinaryColumn` as char)) as `HexText` from `SomeTable`
```

A corresponding Linq query might look like this:

```c#
var hexText = context.SomeTable.Select(t => EF.Functions.Hex((string)t.BinaryColumn)).ToList();
```

mc0re:
With typecast to `string` C# compiler complained: `Cannot convert type 'byte[]' to 'string'.`
Without the typecast the compiler was happy, but runtime complained:
`System.InvalidOperationException: An exception was thrown while attempting to evaluate the LINQ query parameter expression 'EF.Functions.Hex(value(Ncps.Entities.CourseRepository+<>c__DisplayClass12_0).hash)'. ---> System.InvalidOperationException: The specified expression does not have the correct Type.`

lauxjpn:
@mc0re I just pushed #1124 that adds support for `byte[]`.

mc0re:
Good news, I will wait for the next nightly build.

lauxjpn:
@mc0re Should be available via nightly build now.

mc0re:
Build 3.2.0-preview.20362.5
Same exception:
```
System.InvalidOperationException: An exception was thrown while attempting to evaluate the LINQ query parameter expression 'EF.Functions.Hex(value(Ncps.Entities.CourseRepository+<>c__DisplayClass12_0).hash)'. ---> System.InvalidOperationException: The specified expression does not have the correct Type.
```

lauxjpn:
@mc0re What is the LINQ query you are running? Have you tried the sample code I posted?

mc0re:
Query:
```
...
.Where(c => c.Files.Select(f => EF.Functions.Hex(f.Hash)).Contains(EF.Functions.Hex(hash)))
```

lauxjpn:
Thanks for testing. I was able to reproduce the issue. It is not related to the SQL translation, but had to do with the `EF.Functions.Hex(hash)` part, that EF Core wants to executed client side because it does not reference anything database column.

However, the client side implementation of `Hex()` did not yet support `byte[]`.

I pushed #1125, that adds the missing client side support for that.

mc0re:
Everything works now, both in-memory and on MariaDb. Thanks!

lauxjpn:
Hopefully this week. The release is only working on the JSON functionality. Looks good so far though.

PR #1126, opened 2020-07-14
Limit MySqlConnector to 0.x versions

bgrainger:
MySqlConnector 1.0 introduces a breaking namespace change; see https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1103#issuecomment-658277263.

This should prevent clients from upgrading beyond a version that Pomelo supports.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -29,7 +29,7 @@
     <MicrosoftNETCoreAppInternalPackageVersion>3.1.3-servicing.20128.1</MicrosoftNETCoreAppInternalPackageVersion>
     <!-- EFCore.MySql Dependencies -->
     <MicrosoftEntityFrameworkCoreRelationalVersion>3.1.3</MicrosoftEntityFrameworkCoreRelationalVersion>
-    <MySqlConnectorVersion>0.69.2</MySqlConnectorVersion>
+    <MySqlConnectorVersion>[0.69.6, 1.0.0)</MySqlConnectorVersion>
     <PomeloJsonObjectVersion>2.2.1</PomeloJsonObjectVersion>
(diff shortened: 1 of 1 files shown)


bgrainger:
Can this ship as 3.1.2 (with only this change from 3.1.1) to help fix potential dependency problems for users sooner? 3.2.0 looks like it still has a number of open issues and might be a while before it's ready?

lauxjpn:
Let's merge it with the `master` branch. I will cherry-pick it then into the `3.1-maint` branch and publish a `3.1.2` release.

dalton5:
Hi,

I use .net core 2.2 and entity framework 2.

When I don't install the package MySql Hangfire everything works.

When I instakll it I have the following error:

image


Do you know how I can fix the issue?

I can't upgrade now to .net core 3.x

lauxjpn:
Use a version of MySqlConnector below `1.0.0`.  The highest currently available version below `1.0.0` would be `0.69.8`.

bgrainger:
> When I don't install the package MySql Hangfire everything works.
> When I instakll it I have the following error:

I don't know which MySQL Hangfire NuGet package you're using, but you may need to downgrade it to a version that doesn't require MySqlConnector 1.0.0 or later.

bgrainger:
If you're using Hangfire.MySqlStorage then downgrade it to 2.0.2. https://www.nuget.org/packages/Hangfire.MySqlStorage/2.0.2

dalton5:
Thanks @bgrainger  and @lauxjpn .

Downgrade Hangfire.MySqlStorage to 2.0.2 did the trick!
Thanks,

PR #1144, opened 2020-08-17
Use case-insensitive comparison on table names when scaffolding foreign key constraints [copy of original 1017]

lauxjpn:
Since it looks like @dhirensham abandoned #1017 and we currently have a related [question](https://stackoverflow.com/questions/63440367/referenced-table-is-not-in-dictionary/63448835) on StackOverflow, I took the liberty to apply the requested changes myself.

Because the [original PR](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/pull/1017) did not `allow edits be maintainers`, I had to duplicate it.


dhirensham:
@lauxjpn apologies for losing track of this. Thank you for completing this for me, much appreciated.

lauxjpn:
@dhirensham No problem, thanks for reporting and preparing this in the first place!

Issue #1102, opened 2020-06-13
Implement full/extended JSON support


lauxjpn:
As the last major feature before `5.0`, we are going to implement proper JSON support, where either `JsonDocument`, `JsonElement`, `System.String` or custom POCOs can be used in conjunction with the `json` column type.

We will support both stacks, `System.Text.Json` and `Newtonsoft.Json`.

This issue is related to #693 and #598.

mguinness:
Thanks Laurents, appreciate your efforts on this.  @roji Since you've already implemented JSON support for PostgreSQL do you have any pointers and has any design work happened for SQL Server?

roji:
I had hoped we'd be able to deliver SQL Server support for 5.0 (https://github.com/dotnet/efcore/issues/4021), but that's unfortunately not going to happen - too much going on... I definitely think it's going to be a priority item for 6.0.

@lauxjpn it would be great to know more about your plans for the MySQL provider, when those plans become clear :) Ideally we'd manage to align support at least to some extent across providers. If you're curious to see what I did for PostgreSQL, [here's a post on that](http://www.roji.org/efcore-pg-advanced-json) and [here are the docs](http://www.npgsql.org/efcore/mapping/json.html?tabs=data-annotations%2Cpoco). I decided to only support System.Text.Json for now (among other things because Newtonsoft.Json requires adding the dependency to the provider). I implemented both strongly-typed POCO support and weakly-typed DOM support, including some nice query features (i.e. get all documents where a certain property inside the JSON equals something). If you want to implement that kind of thing I'd be happy to help!

lauxjpn:
@roji Thanks for the post, I wasn't aware of that!
I already took ... I mean "borrowed" your work as the baseline for Pomelo, that I want to support first as much as possible (I am about half way done at the moment). After that, it might be time to take a look at what interesting stuff MySQL and MariaDB could bring to the table (probably at least a couple of beers).
As for the two stack support, the current plan is to have two stack specific packages (e.g. `Pomelo.EntityFrameworkCore.MySql.Json.Newtonsoft`), and either an additional package for the common (stack independent) JSON support (like `Pomelo.EntityFrameworkCore.MySql.Json`), or just integrated common JSON support in the main provider package that only works when a provider specific package has been added (for the needed serialization).

roji:
Definitely a couple of beers :) That's starting to be even possible...

[...]

Issue #935, opened 2019-11-11
Update wiki for 3.0.0


lauxjpn:
We should give a quick overview about new features and breaking changes and might also want to mention the top 5 issues that are being reported here, but that are actually just expected behavior changes due to the new query pipeline in EF Core 3.0.0.

Issue #522, opened 2018-03-31
Call for Contributors


caleblloyd:
`Pomelo.EntityFrameworkCore.MySql` is looking for core contributors to help with to help with maintenance and upgrades for the MySql Entity Framework Core Driver to track [EF Core Upstream Version 2.1](https://github.com/aspnet/entityframeworkcore)  and beyond (see issue #514)

I have worked on the previous few minor and major upgrades, and they can take a fair amount of work.  I am no longer using EF at my day job and do not have as much capacity as in times past to contribute.  I can say that it is very fun and rewarding work, and that the EF Core team at Microsoft are amazing folks and responsive when I get completely stuck.  I'm glad to still participate in code reviews and may try to contribute more as I find time.

EF Core performs upstream major, minor, or patch releases that can require re-work of provider code approximately every 6 months.  An ideal core contributor would use both Entity Framework Core and MySQL often and have the time to spend the time interfacing with the EF Upstream team to understand upcoming changes.  That being said, all contributions are welcome.

Thanks for everyone's support over the 1.5 years that I've been working on the project, I look forward to seeing what the community continues to do!

cc @yukozh @mguinness

ngochivinh:
Hi @caleblloyd ,
I'm a software engineer, I want to be a contributor to this project. How can I join the project?
Thank you.
email: vnngochivinh@gmail.com

mguinness:
@caleblloyd @yukozh Does joining [Open Collective](https://opencollective.com/faq) make any sense for this project?  It could be a way to compensate contributors for their time and effort working on the project.

They let any open source project with at least 100 stars on GitHub join under their umbrella where they will act as a host organization so that you can start collecting donations (both one-off and recurring) and paying expenses without the headache of setting up a 501(c)(3) and a formal structure. You just add a bunch of trusted admins and you can manage expenses and invoices. Open Collective holds the money for you.

caleblloyd:
> Does joining Open Collective make any sense for this project?

I don't think it makes a lot of sense at this point, there are currently no expenses associated with this project that I know of.  GitHub, TravisCI, AppVeyor, and MyGet are all free for open source projects.  I like the idea of contributors supporting the project either as personal contributions or companies contributing developer time if their business depends on the project.

VictorHaine:
I'm starting to use Pomelo.EF.MySql and when I get on track maybe I can contribute. Let's see!

T-rav:
[...]

Issue #14, opened 2016-07-14
MySQL 5.7 Json Field Type Is Available Now


Kagamine:
### Getting Started

① Adding pomelo myget feed into your NuGet.config which located in your solution root.

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
    <add key="Pomelo" value="https://www.myget.org/F/pomelo/api/v2/" />
  </packageSources>
  <disabledPackageSources />
</configuration>
```

② Add `Pomelo.Data.MySql` and `Pomelo.EntityFrameworkCore.MySql` into your project.json. The versions of them are `1.0.0`.

③ If you have already installed the pomelo packages: `Pomelo.Data.MySql` and `Pomelo.EntityFrameworkCore.MySql`, please remove them and restore again(`dotnet restore --no-cache`).  The packages are located in `C:\Users\YOURNAME\.nuget\packages`.

④ To define json field in model with `System.JsonObject<T>` will store this field as a json column.

### Sample

``` C#
using System;
using System.Linq;
using System.ComponentModel.DataAnnotations;
using System.ComponentModel.DataAnnotations.Schema;
using Microsoft.EntityFrameworkCore;

namespace MySqlTest
{
    public class Blog
    {
        public Guid Id { get; set; }

        [MaxLength(32)]
        public string Title { get; set; }

        public string Content { get; set; }

        public JsonObject<string[]> Tags { get; set; } // Json storage
    }

    public class MyContext : DbContext
    {
        public DbSet<Blog> Blogs { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
            => optionsBuilder
                .UseMySql(@"Server=localhost;database=ef;uid=root;pwd=19931101;");
    }

    public class Program
    {
        public static void Main()
        {
            using (var context = new MyContext())
            {
                // Create database
                context.Database.EnsureCreated();

                // Init sample data
                var blog1 = new Blog {
                    Title = "Title #1",
                    Tags = new string[] { "ASP.NET Core", "MySQL", "Pomelo" }
                };
                context.Add(blog1);
                var blog2 = new Blog
                {
                    Title = "Title #2",
                    Tags = new string[] { "ASP.NET Core", "MySQL" }
                };
                context.Add(blog2);
                context.SaveChanges();

                // Detect changes test
[...]

PR #1159, opened 2020-09-09
Implement change tracking support for JSON mapped properties

### LINKED ISSUE #1140: JsonObject property setters do not track changes in array elements
### Steps to reproduce
Copy paste the [IceCreames example](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/14#issuecomment-578083962).
Add and change `Program.Main()` to below.

```csharp
using (var context = new Context())
{
    context.Database.EnsureDeleted();
    context.Database.EnsureCreated();

    context.IceCreams.AddRange(
        new IceCream
        {
            Name = "Vanilla",
            Energy = new Energy
            {
                Kilojoules = 866.0,
                Kilocalories = 207.0,
            },
            Comments = new[]
            {
                    "First!",
                    "Delicios!"
            },
        },
        new IceCream
        {
            Name = "Chocolate",
            Energy = new Energy
            {
                Kilojoules = 904.0,
                Kilocalories = 216.0,
            },
            Comments = new[]
            {
                    "My husband linkes this one a lot."
            },
        });

    context.SaveChanges();
}

using (var context = new Context())
{
    IceCream iceCream = context.IceCreams.Where(e => e.IceCreamId == 1).First();
    //iceCream.Comments.Object[0] = "Test update.";
    //iceCream.Comments.Object = new string[] { "Test update.", "Second line." };
    //iceCream.Comments.Json = "[ 'Test update.', 'Second line.' ]";
    iceCream.Comments = new string[] { "Test update.", "Second line." };
    iceCream.Comments = "[ 'Test update.', 'Second line.' ]";
    context.SaveChanges();
}

using (var context = new Context())
{
    var result = context.IceCreams
        .OrderBy(e => e.IceCreamId)
        .ToList();

    Debug.Assert(result.Count == 2);

    Debug.Assert(result[0].Name == "Vanilla");
    Debug.Assert(result[0].Energy.Object.Kilojoules == 866.0);
    Debug.Assert(result[0].Comments.Object.Length == 2);
    Debug.Assert(result[0].Comments.Object[0] == "Test update.");
}
```

### The issue
Assertion failed with below commented cases.

```csharp
//iceCream.Comments.Object[0] = "Test update.";
//iceCream.Comments.Object = new string[] { "Test update.", "Second line." };
//iceCream.Comments.Json = "[ 'Test update.', 'Second line.' ]";
iceCream.Comments = new string[] { "Test update.", "Second line." };
iceCream.Comments = "[ 'Test update.', 'Second line.' ]";
```

Only assertion succeed with implicit operators(last two ones).

I can understand first one fails. It is almost impossible to track changes and other systems also do not consider such like things.

But it is confusing second and third one fail which are accessing property setters. I think, removing the setters or implement as same with the implicit operators is better.

### Further technical details

MySQL version: MariaDB 10.3
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 3.12
Microsoft.AspNetCore.App version: .NET Core 3.1 Console Application

lauxjpn:
While the bug was most apparent with POCO array elements and POCO properties in general, this fix applies to all JSON mapped properties (string, DOM and POCO).

Relates to #1102, #1157 and #1158
Fixes #1140


src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Pomelo.EntityFrameworkCore.MySql.Storage.ValueComparison.Internal
+{
+    public class JsonValueComparer<T> : ValueComparer<T>
+    {
+        public JsonValueComparer(ValueConverter valueConverter)
+            : base(
+                (left, right) => object.Equals(valueConverter.ConvertToProvider(left), valueConverter.ConvertToProvider(right)),
+                value => value.GetHashCode(),
+                value => (T)valueConverter.ConvertFromProvider(valueConverter.ConvertToProvider(value)))
```
AndriySvyryd, line 12: The snapshot lambda is invoked for every property when the entity is tracked or is saved. Making it faster would result in noticeable perf improvements.

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.EntityFrameworkCore.ChangeTracking;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Storage.ValueComparison.Internal
+{
+    public class JsonValueComparer<T> : ValueComparer<T>
+    {
+        public JsonValueComparer(ValueConverter valueConverter)
+            : base(
+                (left, right) => object.Equals(valueConverter.ConvertToProvider(left), valueConverter.ConvertToProvider(right)),
```
AndriySvyryd, line 10: Equals lambda is also on the hot path.

AndriySvyryd (approved):

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
lauxjpn: It could be improved for _some_ scenarios as far as I can see:
* Newtonsoft.Json if `T` is a DOM class: Use [JTokenEqualityComparer](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_JTokenEqualityComparer.htm)
* Any Json-Library if `T` is `string`: Just compare the strings as is.

This would however do nothing for the most common case, which should be to use POCOs.

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
lauxjpn: The only general way I could think about speeding this up was to traverse the object graphs ourselves in parallel, so we can compare them node by node and possibly break early, if they diverge. This would however be significantly more complex then just letting the JSON library do the work and comparing the result.

For Newtonsoft.Json, if `T` is a DOM class, then [JTokenEqualityComparer](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Linq_JTokenEqualityComparer.htm) could be used.

lauxjpn:
However, the moment we at least don't process JSON mappings that are to a `T` of `string`, users would have at least an implicit way to improve performance by using `string` instead of a POCO.

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
AndriySvyryd: You could detect whether it implements `IEquatable<T>` or overrides `Equals` and use that.

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
AndriySvyryd: You could check for `ICloneable` implementation, `JToken` has `DeepClone` and I believe JsonDocument is immutable, so same value can be returned

lauxjpn: > `JToken` has `DeepClone` and I believe `JsonDocument` is immutable

Good points!

> You could check for `ICloneable` implementation

Theoretically true, though the `ICloneable` contract is traditionally not explicit in _how_ the object is being cloned. From the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.icloneable?view=netcore-3.1#notes-to-implementers):

> Because callers of `Clone()` cannot depend on the method performing a predictable cloning operation, we recommend that `ICloneable` not be implemented in public APIs.

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
lauxjpn: > You could detect whether `T` implements `IEquatable<T>` or overrides `Equals` and use that.

Good points!

src/EFCore.MySql/Storage/ValueComparison/Internal/JsonValueComparer.cs
AndriySvyryd: You could also look for `DeepClone`, as long as this is documented it will allow to implement it efficiently for POCOs

lauxjpn: > You could also look for `DeepClone()`, as long as this is documented it will allow to implement it efficiently for POCOs

Could be a better approach. As for `ICloneable`, we might make this opt-in/opt-out.

roji: While JsonDocument is immutable and so there's no need to actually snapshot or do anything beyond reference comparison, JsonElement is a struct, and so that's not really possible... We should consider not allowing to map JsonElement.

lauxjpn: I agree that it makes little sense to snapshot `JsonElement` with the default struct member by member copy in the way `JsonElement` has been implemented. If the `JsonElement._parent` field had a publicly available property, this wouldn't be a problem, since the parent would still be immutable. But for some reason (no idea why) it has no property.

roji: You're right (https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1140#issuecomment-691620821)

lauxjpn:
I added the optimized change tracking support for `System.Text.Json` and `Newtonsoft.Json`.

The `UseMicrosoftJson` and `UseNewtonsoftJson` extension methods now contain an optional `MySqlCommonJsonChangeTrackingOptions` parameter, that can be used to specify the global default change tracking options used for all JSON column mapped properties. This should help immensely with the discoverability of the change tracking options. The default value is `RootPropertyOnly`.

Currently, we only have one value comparer for each JSON support library, the handles the expression construction for everything JSON. We can split it up later if necessary (like we have for the JSON value converters). This should not result in backwards compatibility issue, because we marked the value comparers (as well as the value converters) as _internal_.

The value comparers (and converters) are mapped automatically with the global default change tracking options. If a user wants to change the options for specific model properties, we introduced an `UseJsonChangeTrackingOptions` property builder extension method per JSON support library.

So unless users don't explicitly opt into internal behavior, the internals have been abstracted away. For EF Core 6 and its possible JSON out-of-the-box support, this gives us a lot of flexibility. E.g. we could alter the internal in accordance to new requirements and then make them publicly available, while keeping our current interface.

The publicly available `MySqlCommonJsonChangeTrackingOptions` are common combinations of the internal `MySqlJsonChangeTrackingOptions` flags, that finely control how comparison, snapshotting and hashing is handled.

sukhdev22:
Surprisingly, I don't see a single **Sample application** which showcases how to map JSON columns to POCO .NET properties. How to insert, update and query Out-Of-The-Box using this library. Can anyone point to any such resource ?

lauxjpn:
@sukhdev22 The most comprehensive sample code we currently have is probably the one from [#14 (comment)](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/14#issuecomment-725257107).

It demonstrates all three types of JSON support: POCOs, the DOM and strings.

mguinness:
Yes, both documentation and sample code could be improved and are highlighted in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/522 and https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/935.

There is also https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1179 which asks for help in that effort, so if you could contribute it would be welcomed.

PR #1166, opened 2020-09-17
Fix bool indexing optimization, which could lead to incorrect GROUP BY statements

lauxjpn:
Prerequisite for enabling `ONLY_FULL_GROUP_BY` in all tests with #1165.


lauxjpn:
This will be part of the #1165 PR.

Issue #1163, opened 2020-09-16
Spatial Queries


aderrose:
### Steps to reproduce
Using the project code from https://github.com/gavilanch/EFCoreSpatialQueries I've modified the ApplicationDBContext to point to a SQL Server 2016 LocalDB (as per original project), a MySQL 8.0.21 server (via Pomelo) and a MariaDB 10.5.5 server (via Pomelo).

Each time I change server the migrations need deleting, recreating and applying to the database, otherwise no code is modified from original project.

### The issue
The project demonstrates loading the database with spatial data and querying that back limiting the dataset to places within 2000 meters.

The LocalDB database seems to work correctly and returns :

> Agora from Santo Domingo (302.858257326255 meters away)
> Adrian Tropical from Santo Domingo (1188.9001109225335 meters away)

The following two entries can be displayed by removing the 2000 meters where clause:

> Sambil from Santo Domingo (2863.422068459216 meters away)
> Restaurante El Cardenal from Mexito City (3073097.7298800712 meters away)

However, if I run the same code against MySQL 8 I get the following output:

> Adrian Tropical from Santo Domingo (11858806.351341443 meters away)
> Sambil from Santo Domingo (11858886.04280507 meters away)
> Agora from Santo Domingo (11859348.903124722 meters away)
> Restaurante El Cardenal from Mexito City (11884292.875515593 meters away)

All of which are well over 2000 meters and very different to the LocalDB version.

When I run the code against MariaDB 10.5 I get the following:

> Agora from Santo Domingo (0.0027362819628112514 meters away)
> Adrian Tropical from Santo Domingo (0.010865760255958656 meters away)
> Sambil from Santo Domingo (0.027108523012515796 meters away)
> Restaurante El Cardenal from Mexito City (29.211945550922625 meters away)

All of which are tiny distances and also very different.

If I try to reproduce the calculations through HeidiSQL using either ST_DISTANCE or ST_DISTANCE_SPHERE (which MariaDB is missing) I get different values again, 5516 meters with SRID of 4326 and 5425 meters if no SRID is set.

I'm now really confused which values are correct and where the issue lies, to me it looks like it's a database engine issue rather than Pomelo or even MySqlConnector as the same code produces different values.

Ultimately what I'm trying to achieve is a method of calculating distance between latitude and longitude coordinates based on data from GPX files to calculate an overall distance for a route.

Any help or explanation as to what is going on would be gratefully received.

[...]

PR #1174, opened 2020-09-25
Fix (lon, lat) reordering across all supported database servers and versions

lauxjpn:
Fixes https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1163#issuecomment-697925451


mguinness:
Discussed at https://dba.stackexchange.com/a/242004/192409

> That's because on MySQL 8, they did the unthinkable and implemented the stupid method of doing geography coordinates (like MySQL). This broke backwards compatability.

Also looks like [WL#9436](https://dev.mysql.com/worklog/task/?id=9436) introduced the [ST_SwapXY](https://dev.mysql.com/doc/refman/8.0/en/gis-format-conversion-functions.html#function_st-swapxy) function.

Issue #906, opened 2019-11-01
Support `SEQUENCE` for MariaDB


lauxjpn:
In contrast to MySQL, MariaDB introduced `SEQUENCE` in `10.3`.
See the [Knowledge Base](https://mariadb.com/kb/en/library/create-sequence/) for further details.

This would fix #768 and address #509 and #477 for MariaDB.
I think we should consider implementing this for `3.1.0`.

pgiacomo69:
My Pull request solves 2 Issues:
1)Syntax Mismatch: "NOCYCLE" instead of "NOCYCLE"
2)Values Mismatch: Using "Starts with 0" is incompatible with NO MINVALUE (Minvalue has to be set when starting with 0 or less

lauxjpn:
Thanks to @pgiacomo69 for the sequence support in regards to migrations!

PR #1195, opened 2020-10-13
Sequence sql generator

pgiacomo69:
Solves Issue "Support `SEQUENCE` for MariaDB #906"


lauxjpn:
@pgiacomo69 Thanks for contributing. This could be a start to implementing #906 feature.

To properly support sequences in general, including support for queries and scaffolding, there is more work needed (see [Npgsql](https://github.com/npgsql/efcore.pg/search?q=sequence) for example).

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -391,16 +391,10 @@ protected override void Generate(DropSchemaOperation operation, IModel model, Mi
             builder
```
lauxjpn, line 391: The current implementation of [MigrationsSqlGenerator.Generate(CreateSequenceOperation operation, ...)](https://github.com/dotnet/efcore/blob/55e8f0a65e054ab6248474e4755d9e153c0f1c28/src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs#L521) has changed since this method was added (which we shouldn't have done in the first place) to `MySqlMigrationsSqlGenerator`. Therefore, take a look at the current base class implementation first.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -758,7 +752,7 @@ protected virtual string[] GetColumns([NotNull] IEnumerable<IProperty> propertie
                 builder.Append(" NO MAXVALUE");
             }
 
-            builder.Append(operation.IsCyclic ? " CYCLE" : " NO CYCLE");
+            builder.Append(operation.IsCyclic ? " CYCLE" : " NOCYCLE");
```
lauxjpn, line 755: This should be implemented in an overridden [MigrationsSqlGenerator.SequenceOptions()](https://github.com/dotnet/efcore/blob/55e8f0a65e054ab6248474e4755d9e153c0f1c28/src/EFCore.Relational/Migrations/MigrationsSqlGenerator.cs#L1321) implementation.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -391,16 +391,10 @@ protected override void Generate(DropSchemaOperation operation, IModel model, Mi
...
                 .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name, operation.Schema));
-
-            if (operation.ClrType != typeof(long))
+            if (operation.StartValue <= 0)
             {
-                var typeMapping = Dependencies.TypeMappingSource.GetMapping(operation.ClrType);
-
-                builder
-                    .Append(" AS ")
-                    .Append(typeMapping.StoreType);
```
lauxjpn, line 401: Take a look how [Npgsql](https://github.com/npgsql/efcore.pg/blob/795cb6bdf789ae601c2a483178850c930fbc5d4b/src/EFCore.PG/Migrations/NpgsqlMigrationsSqlGenerator.cs#L1312-L1330) implemented the method.
We should probably implement it in a similar way, so we have to maintain as little custom code as possible.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -391,16 +391,10 @@ protected override void Generate(DropSchemaOperation operation, IModel model, Mi
             builder
                 .Append("CREATE SEQUENCE ")
```
lauxjpn, line 392: `CREATE SEQUENCE` is only supported by some MariaDB versions as of now. Therefore, this methods needs to check the server version accordingly. Add the necessary support constructs to [ServerVersion.Support.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/EFCore.MySql/Storage/ServerVersion.Support.cs) and then use them to check that support exists here. For the general idea, take a look at the [Npgsql](https://github.com/npgsql/efcore.pg/blob/795cb6bdf789ae601c2a483178850c930fbc5d4b/src/EFCore.PG/Migrations/NpgsqlMigrationsSqlGenerator.cs#L1312-L1330) implementation (though our syntax is different when in comes to version checks).

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -391,16 +391,10 @@ protected override void Generate(DropSchemaOperation operation, IModel model, Mi
...
                 .Append(Dependencies.SqlGenerationHelper.DelimitIdentifier(operation.Name, operation.Schema));
```
lauxjpn, line 393: We need a tests to verify the behavior. They should be implemented in [MigrationSqlGeneratorMySqlTest.MySql.cs](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.MySql.cs).

Because this feature is supported only by some MariaDB versions, most tests will need to be annotated using `SupportedServerVersionFactAttribute`.
Some might need the opposite, which is `SupportedServerVersionLessThanFactAttribute`.

lauxjpn (changes requested):

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
pgiacomo69: Right, that way is cleaner

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
pgiacomo69: Am I wrong, or it seems that the only way to override this method is to rewrite it as we are doing? I don't see a way to change anything in MigrationCommandListBuilder builder after base method execution.

lauxjpn: Yes, you will basically have to copy the whole `SequenceOptions()` method, change `virtual` to `override` and then change `NO CYCLE` to `CYCLE`. I don't see any reliably shortcuts there.

pgiacomo69: I updated the 2 Methods as we discussed,  Npgsql-like Implementation is better.
I've also implemented same modification on 5.0 branch in my repository, sorry, no time now to implement test cases so I didn't make the pull request. I think that by the end of week I will complete it

pgiacomo69: Do you want that I PR it against 5.0 now, or after implementing test cases?

test/EFCore.MySql.FunctionalTests/TestUtilities/Attributes/SupportedServerVersionFactAttribute.cs
```diff
@@ -17,7 +18,7 @@ private SupportedServerVersionFactAttribute(ServerVersionSupport serverVersionSu
 
             if (!serverVersionSupport.IsSupported(currentVersion) && string.IsNullOrEmpty(Skip))
             {
-                Skip = $"Test is supported only on {serverVersionSupport.SupportedServerVersions} and higher.";
+                Skip = $"Test is supported only on {serverVersionSupport.SupportedServerVersions.Join()} and higher.";
```
lauxjpn, line 21: Nice catch. We should actually be able to use `serverVersionSupport` directly, because its `ToString()` overload should take care of the concatenation.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -385,35 +385,26 @@ protected override void Generate(DropSchemaOperation operation, IModel model, Mi
...
         protected override void Generate(
-            CreateSequenceOperation operation,
-            IModel model,
-            MigrationCommandListBuilder builder)
+            [NotNull] CreateSequenceOperation operation,
+            [CanBeNull] IModel model,
+            [NotNull] MigrationCommandListBuilder builder)
         {
             Check.NotNull(operation, nameof(operation));
-            Check.NotNull(builder, nameof(builder));
```
lauxjpn, line 393: We should throw a `NotSupportedException`, if sequences are not supported. Something similar to:
https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/ca2d02fa2ee38c13c2eb2a56bbc4a0298e89f039/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs#L847-L849
Finally, a supporting test that ensures that an appropriate exception is thrown when an unsupported server version is used, needs to be added as well.

lauxjpn (changes requested):

lauxjpn:
You can ignore the other failing test. As long as your tests are green, we will merge this PR anyway.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
pgiacomo69: Yes, but implementing this, automatically breaks tests. 
When test are executed, the annotation [SupportedServerVersionFact(ServerVersion.SequencesSupportKey)] works as expected, since in Config.json I put "10.5.4-MariaDB" but _options.ServerVersion from MySqlMigrationsSqlGenerator  class has the value of 8.0.21-mysql (default), maybe it was never initialized?

lauxjpn: There is a `Generate` method I added just recently, that allows to specify a `MySqlDbContextOptionsBuilder`, that allows you to configure the server version:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/b31b2acd4225cb927600282dff7f97d4eacddb79/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.cs#L1012-L1024

Used e.g. here:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/b31b2acd4225cb927600282dff7f97d4eacddb79/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.MySql.cs#L141-L169

I merge it tomorrow into the `5.0-wip` branch, so you can use it.

lauxjpn:
@pgiacomo69 We won't be able to merge your PR until you rebase it on the top of the `master` branch.

pgiacomo69:
Hi, sorry for late answer. Today I'm trying to rebase this PR, almost done. 
In late version I was able to get the minimum server version for a feature to work, example:
```
$"Error in Restart Sequence {operation.Name}: Sequences are not supported." +
                    $"Sequences are  supported only from  {ServerVersion.Supports. GetSupport(ServerVersion.SequencesSupportKey)}.");

```
I see that ServerVersion class now is changed, I implemented "Sequences" property in "ServerVersionSupport" class and overrided in "MariaDbServerVersionSupport "but  it seems that is not possible to get that information anymore, am I right?

test/EFCore.MySql.FunctionalTests/MySqlMigrationsSqlGeneratorTest.cs
```diff
@@ -1071,6 +1071,109 @@ PRIMARY KEY (`Name`, `Brand`(20))
                 ignoreLineEndingDifferences: true);
         }
 
+       
+        [SupportedServerVersionCondition(nameof(ServerVersionSupport.Sequences))]
```
lauxjpn, line 1075: All tests need to be marked with the `[ConditionalFact]` attribute.

test/EFCore.MySql.FunctionalTests/MySqlMigrationsSqlGeneratorTest.cs
```diff
@@ -1071,6 +1071,109 @@ PRIMARY KEY (`Name`, `Brand`(20))
...
+               @"CREATE SEQUENCE `MySequence` START WITH 10 INCREMENT BY 1 MINVALUE 10 MAXVALUE 20 NOCYCLE;" + EOL,
+               Sql,
+               ignoreLineEndingDifferences: true);
+        }
+
+        [SupportedServerVersionCondition(nameof(ServerVersionSupport.Sequences))]
+        public virtual void CreateSequenceOperation_with_minValue_and_maxValue_not_long()
+        {
+            throw(new Exception("MariaDb Sequences can be only long"));
+        }
```
lauxjpn, line 1139: Why does this test exist? It doesn't do anything but will just always fail.

test/EFCore.MySql.FunctionalTests/MySqlMigrationsSqlGeneratorTest.cs
```diff
@@ -1071,6 +1071,109 @@ PRIMARY KEY (`Name`, `Brand`(20))
...
+
+        [SupportedServerVersionCondition(nameof(ServerVersionSupport.Sequences))]
+        public virtual void CreateSequenceOperation_without_minValue_and_maxValue()
+        {
+
```
lauxjpn, line 1144: Empty line.

test/EFCore.MySql.FunctionalTests/MySqlMigrationsSqlGeneratorTest.cs
```diff
@@ -1071,6 +1071,109 @@ PRIMARY KEY (`Name`, `Brand`(20))
...
+            Assert.Equal(
+               @"CREATE SEQUENCE `MySequence` START WITH 1 INCREMENT BY 1 NO MINVALUE NO MAXVALUE NOCYCLE;" + EOL,
+               Sql,
+               ignoreLineEndingDifferences: true);
+        }
+
+        [SupportedServerVersionCondition(nameof(ServerVersionSupport.Sequences))]
+        public virtual void DropSequenceOperation()
+        {
+
```
lauxjpn, line 1163: Empty line.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
lauxjpn: The commented part should be removed.

The message should also be changed to something more explicit, that hints at the fact that this is not a general error, but has to do with the currently used server version:

```
$"Cannot create sequence '{operation.Name}' because sequences are not supported in server version {_options.ServerVersion}."
```

The same would apply to the exception in `RestartSequenceOperation`.

lauxjpn (changes requested):
I gave this another review. A couple of smaller issues, and the major issue that the tests are not being run remain.

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
lauxjpn: This applies to the following lines:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/9232252659c39ad4016e4ae8696da807c001e227/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs#L400-L402

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/9232252659c39ad4016e4ae8696da807c001e227/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs#L281-L283

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -279,8 +279,7 @@ protected override void Generate(AlterTableOperation operation, IModel model, Mi
             if (!_options.ServerVersion.Supports.Sequences)
             {
                 throw new InvalidOperationException(
-                    $"Error in Restart Sequence {operation.Name}: Sequences are not supported." /* +
-                    // $"Sequences are  supported only from  {ServerVersion.Supports.GetSupport(ServerVersion.SequencesSupportKey)}." */); 
+                    $"Cannot create sequence '{operation.Name}' because sequences are not supported in server version {_options.ServerVersion}.");
```
lauxjpn, line 282: This should be `Cannot restart sequence ...`.

lauxjpn (changes requested):
Only one small issue left.

lauxjpn (approved):

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
pgiacomo69: Done, sorry for my sloppiness and thank you for your patience

lauxjpn:
@pgiacomo69 Thanks for your contribution!

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
lauxjpn: No worries, all good!

pgiacomo69:
Thank You again for your patience, this was an occasion to learn git and github, hope to see soon the package on nuget, I'm actively using it.

lauxjpn:
Should be out by the end of the week, but it is now already part of the latest [nightly build](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#nightly-builds).

PR #1218, opened 2020-11-01
Check and reimplement nullability and argument null propagation for all SQL functions

lauxjpn:
While I revisited all SQL function translations and their argument null propagation, I seized the opportunity to document them all in an Excel sheet, so we can add them to the wiki later, so users can lookup which properties/method calls are being translated to SQL. I'll attach it here, so it cannot get lost:

[Translations_20201031_01.xlsx](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/files/5471892/Translations_20201031_01.xlsx)


mguinness:
I took a first stab at creating a [Translations](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/wiki/Translations) wiki page, please feel free to change where necessary.

lauxjpn:
@mguinness Awesome!

PR #1244, opened 2020-11-14
Remove CharSetBehavior

lauxjpn:
Character sets can be managed by using the `HasCharSet()` extension method.
So the `CharSetBehavior` option is now obsolete.

Also introduces the `LimitKeyedOrIndexedStringColumnLength()` option, to control whether the default length of string columns that participate in any kind of index should be limited to half the size of the max key length value, if they have not been explicitly specified with a column length or index prefix length. This has been the default value in the past and continues to be the default, but this option allows users to disable this behavior (in which case the column will take the usual string column default length, with is currently 4000 for Unicode columns). Finer control is always possible using the `PrefixLength()` setting on the index in the model definition.


gitown:
@lauxjpn Hi,
Any update regarding this PR?

lauxjpn:
@gitown I am working on it over the weekend.

Issue #678, opened 2018-09-03
Migration: Changing an index that is part of an FK relationship fails


LordMike:
### Steps to reproduce

1. Create two types:
  * A: int Id, int BId; use an AK on A.BId
  * B: int Id;
2. Create an FK relationsip. A.BId -> B.Id
3. Create Migration
4. Alter A so that it no longer uses an AK but rather a Unique Index (I was troubleshooting the exact same issue as in #90)
5. Create migration

### The issue
In the final result, Pomelo should have generated a "DROP CONSTRAINT" statement (which is wrong - but nevermind that), without first removing the FK constraint. The DROP will fail, with mysql stating that it can't be done.

I've workarounded it by dropping the FK, removing the AK, adding a unique index, then re-creating the FK.

Attached here, [ConstraintsIssue.zip](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/files/2346374/ConstraintsIssue.zip), is a solution with the issue.

Using `dotnet ef migrations script`, it will create the following:

```mysql
CREATE TABLE `__EFMigrationsHistory` (
    `MigrationId` VARCHAR(95) NOT NULL,
    `ProductVersion` VARCHAR(32) NOT NULL,
    CONSTRAINT `PK___EFMigrationsHistory` PRIMARY KEY (`MigrationId`)
);

CREATE TABLE `Bs` (
    `Id` INT NOT NULL AUTO_INCREMENT,
    CONSTRAINT `PK_Bs` PRIMARY KEY (`Id`)
);

CREATE TABLE `As` (
    `Id` INT NOT NULL AUTO_INCREMENT,
    `BId` INT NOT NULL,
    CONSTRAINT `PK_As` PRIMARY KEY (`Id`),
    CONSTRAINT `AK_As_BId` UNIQUE (`BId`),
    CONSTRAINT `FK_As_Bs_BId` FOREIGN KEY (`BId`) REFERENCES `Bs` (`Id`) ON DELETE CASCADE
);

INSERT INTO `__EFMigrationsHistory` (`MigrationId`, `ProductVersion`)
VALUES ('20180903184926_Initial', '2.1.2-rtm-30932');

-- ALTER TABLE `As` DROP CONSTRAINT `AK_As_BId`; -- Ignore this error for now
ALTER TABLE `As` DROP INDEX `AK_As_BId`;

[...]

PR #1282, opened 2020-12-29
Redesign foreign key recreation when dropping primary keys or unique constraints

### LINKED ISSUE #1280: Migration for changing Key Column (along with foreign keys attached to it) fails on removing/adding fk constraints
I'm sorry that I'm reporting two problems in a single issue. I was trying to reproduce a problem but I faced another one which was stopping me to reproduce the main problem. So I'm going to explain both of them

---

### Steps to reproduce
here's my example [code](https://github.com/ahmadalli/PomeloMysqlFKTest)

### The issue

If you run the sample project I provided (it's runnable on docker-compose), you'll face this exception when `StartupHostedService` tries to run  `await dbContext.Database.MigrateAsync(cancellationToken);`.

Exception message:

this is the exception you get when you run the [project](https://github.com/ahmadalli/PomeloMysqlFKTest)

```
MySqlConnector.MySqlException: 'Can't DROP 'FK_Children_Parents_ParentId'; check that column/key exists
```

Stack trace:
```
   at MySqlConnector.MySqlDataReader.ActivateResultSet(CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 131
   at MySqlConnector.MySqlDataReader.CreateAsync(CommandListPosition commandListPosition, ICommandPayloadCreator payloadCreator, IDictionary`2 cachedProcedures, IMySqlCommand command, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlDataReader.cs:line 436
   at MySqlConnector.Core.CommandExecutor.ExecuteReaderAsync(IReadOnlyList`1 commands, ICommandPayloadCreator payloadCreator, CommandBehavior behavior, IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/Core/CommandExecutor.cs:line 60
   at MySqlConnector.MySqlCommand.ExecuteNonQueryAsync(IOBehavior ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlCommand.cs:line 264
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteNonQueryAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteNonQueryAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.RelationalCommand.ExecuteNonQueryAsync(RelationalCommandParameterObject parameterObject, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.MigrationCommandExecutor.ExecuteNonQueryAsync(IEnumerable`1 migrationCommands, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.MigrationCommandExecutor.ExecuteNonQueryAsync(IEnumerable`1 migrationCommands, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.MigrationCommandExecutor.ExecuteNonQueryAsync(IEnumerable`1 migrationCommands, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.MigrationCommandExecutor.ExecuteNonQueryAsync(IEnumerable`1 migrationCommands, IRelationalConnection connection, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator.MigrateAsync(String targetMigration, CancellationToken cancellationToken)
   at FKTest.Services.HostedServices.StartupHostedService.StartAsync(CancellationToken cancellationToken) in E:\tmp\FKTest\FKTest\Services\HostedServices\StartupHostedService.cs:line 35
   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.RunAsync(IHost host, CancellationToken token)
   at Microsoft.Extensions.Hosting.HostingAbstractionsHostExtensions.Run(IHost host)
   at FKTest.Program.Main(String[] args) in E:\tmp\FKTest\FKTest\Program.cs:line 15
```

I guess it's happening when ef is trying to run this part of the migration

https://github.com/ahmadalli/PomeloMysqlFKTest/blob/e432af74d8ad3ef0c964dc376a6a3698ea0993e4/FKTest/Migrations/20201227210928_ChangeIdColumn.cs#L17-L19

and it makes sense since `FK_Children_Parents_ParentId` has not been created yet, The migration should delete `FK_Children_Parents_ParentCode` instead.

### The Main Problem

I was facing another issue on my project and I wanted to report that (with similar section on migration code) but I faced this issue first. The issue I'm facing on another project is that the migration fails because it tries to run this query which obviously fails because there's no `ParentId` column in `Parents` table:

```sql
ALTER TABLE `Children` ADD CONSTRAINT `FK_Children_Parents_ParentId` FOREIGN KEY (`ParentId`) REFERENCES `Parents` (`ParentId`) ON DELETE CASCADE;
```

I was looking for the code that generates this query but I couldn't find it but I found the code that's supposed to test it:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/c8647fc801397185faa429878d361c949826cf26/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.MySql.cs#L49-L54

But I think these lines doesn't cover all the corner cases since it the `LicensePlateNumber` is the name of foreign key column and also the name of primary column on the reference table.

### Further technical details

MySQL version: 5.7
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 5.0.0-alpha.2
Microsoft.AspNetCore.App version: 5.0.1

lauxjpn:
This PR changes the fix for #678. 

Previously, we automatically removed all foreign keys, then dropped the primary key or unique constraint and finally readded all previously droppend foreign keys.

The problem with that is, that when the primary key is changed in the same migration as one of the tables foreign keys, we might drop a non-existing foreign key or EF Core might have already issued an operation to drop the foreign key.

Therefore, we do not automatically drop foreign keys anymore at the moment. Instead, we allow users to opt-in the foreign key drop behavior using a `DropPrimaryKey()` and `DropUniqueConstraint()` overload (actually an extension method) with an additional parameter.

Also adds a lot of tests.

Fixes #1280


ahmadalli:
I don't understand why this was happening on the first place (why foreign keys of a table should be recreated if that table's PK was changed). If I understand correctly, this PR makes it optional to do so. If that's true, I think we also need to have proper documentations on the case (and if possible, issue a warning when migration is created) and this might break backward compatibility if the default behavior changes from dropping FKs to not dropping them

lauxjpn:
> I don't understand why this was happening on the first place (why foreign keys of a table should be recreated if that table's PK was changed).

See #678 (especially https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/678#issuecomment-579823838).

In you particular case, you changed both, the foreign key and the primary key. The foreign key was dropped first by EF Core. Then the primary key drop was executed. This operation tries to drop all foreign keys (including the already dropped one), but relying on the ones that it can get from the target snapshot (which in your case does not contain the already dropped foreign key, but the new foreign key that will be added later in the same migration). Therefore, the result was to drop a foreign key that did not exist yet.

> I think we also need to [...] issue a warning when migration is created

That was already the case for your migration:

```
PS E:\Sources\PomeloIssues\Issue1280\PomeloMysqlFKTest\FKTest> dotnet ef migrations add ChangeIdColumn
Build started...
Build succeeded.
info: Microsoft.EntityFrameworkCore.Infrastructure[10403]
      Entity Framework Core 5.0.1 initialized 'FkDbContext' using provider 'Pomelo.EntityFrameworkCore.MySql' with options: DetailedErrorsEnabled MaxPoolSize=128 ServerVersion 5.7-mysql
An operation was scaffolded that may result in the loss of data. Please review the migration for accuracy.
Done. To undo this action, use 'ef migrations remove'
```

> this might break backward compatibility if the default behavior changes from dropping FKs to not dropping them

I added the `breaking-change` tag. I also already commented on the original issue thread #678.

Since migrations should always be scripted, and those scripts should be tested first, this should not be an issue.
In the long run, we will fix the issue by dropping an recreating the foreign keys in SQL, instead of relying on the target snapshot (which might be incorrect when multiple changes happen in one migration).

PR #1295, opened 2021-01-15
workaround/issue1293_extended

### LINKED ISSUE #1293: Invalid sql statement is generated by the library when integrated with odata / MySQL 5.7+ returns NULL for constant values in LEFT JOINs if statement contains ORDER BY
### Steps to reproduce
Steps to reproduce were completely described in the separate issue created for OData/WebApi here: https://github.com/OData/WebApi/issues/2124

Please find the example where the error can be reproduced here: https://github.com/huiali/odata-sample

### The issue
```
OData URL-->GET
https://localhost:5002/oq/A?$expand=B($expand=C)-->
ERROR --> Value cannot be null. (Parameter 'key')
```

Using OData, expand inside another expand, or filter inside expand the error appears.

Investigations in the separate thread (https://github.com/OData/WebApi/issues/2124) showed that the issue is with the sql statement that is generated by the Pomelo.EntityFrameworkCore.MySql lib when OData is executing expand inside another expand

### Further technical details
Pomelo.EntityFrameworkCore.MySql version: 3.2.4
Microsoft.AspNetCore.App version: 3.1
Microsoft.AspNetCore.OData version: 7.5.0

OlehGaras:
Extended the fix to be applied to the parameter expressions as well

Fixes #1293


lauxjpn (changes requested):
@OlehGaras Are you sure this is necessary? Is far as I am aware of, only constant values trigger the bug.
If you think, that also parameters can trigger the bug, please add a test to prove this behavior.

I just added a test case called `MySqlBug96947.Bug_exists_in_MySql57_or_higher` to the `EFCore.MySql.Tests` project (you need to copy your `config.json` from `EFCore.MySql.FunctionalTests` over), that proves the constant value bug (see #1298).
Just add your test case to the `MySqlBug96947` class as well, that proves the bug is also being triggered by parameters.

OlehGaras:
hi @lauxjpn , 
I have added test fixture proving  that the bug also exists for parameters.
I also renamed file with you test.

Please take a look

OlehGaras:
@lauxjpn , did you have a chance to take a look ?

lauxjpn:
@OlehGaras No need for the new test class and fixture. I already had the additional test case ready and also some additional optimization, so that the workaround is only applied if an `ORDER BY` statement is being used.

lauxjpn (approved):

lauxjpn:
@OlehGaras Thank you for your contribution!

OlehGaras:
@lauxjpn , thank you as well for your help! 
So the fix will be available at nightly builds as prereleases, right ?

lauxjpn:
> So the fix will be available at nightly builds as prereleases, right ?

@OlehGaras Yes, it is already merged with the master branch and the CI for the commit has completed, so it is already available in the latest nightly build.

Issue #1205, opened 2020-10-21
To input a sytem type to Annotation() function instead of MySqlValueGenerationStrategy type


JeffreySu:
### Steps to reproduce

The code grenerated by Add-Migration in console:

``` C#
    public partial class AddConfig : Migration
    {
        protected override void Up(MigrationBuilder migrationBuilder)
        {
            migrationBuilder.CreateTable(
                name: "MuTable",
                columns: table => new
                {
                    Id = table.Column<int>(nullable: false)
                        .Annotation("MySql:ValueGenerationStrategy", MySqlValueGenerationStrategy.IdentityColumn),
                    //...

                }
        }
}
```

### The issue
Sometimes while we want to just create a migration file, and store it in a CLEAN assembly, we need a CLEAN dependency to the provider. 

Here are two practice:

The same setting for SqlServer is:
``` C#
Id = table.Column<int>(nullable: false).Annotation("SqlServer:Identity", "1, 1"),
```
The same setting for SQLite is:
``` C#
Id = table.Column<int>(nullable: false).Annotation("Sqlite:Autoincrement", true),
```

They just support String (`"1, 1"`) and Bool (`true`) type, they don't have to depend to `Microsoft.EntityFrameworkCore.SqlServer` or `Microsoft.EntityFrameworkCore.Sqlite`

So if we can support a String(like `"IdentityColumn"`) instead of an Enum type (`MySqlValueGenerationStrategy.IdentityColumn`) in the migration file, it will get better scalability.

### Further technical details

MySQL version: 8.0
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 3.1.2
Microsoft.AspNetCore.App version: 3.1

Other details about my project setup:

lauxjpn:
We have planned to update the whole value generation and identity code for 5.0. We will take a look at this issue in the process.

JeffreySu:
Thanks @lauxjpn , when will v5.0 be released?

mguinness:
Watch issue https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1088 for latest updates on 5.0 release.

JeffreySu:
> We have planned to update the whole value generation and identity code for 5.0. We will take a look at this issue in the process.
> 
> In the meantime, you should be able to just store the underlying `int` value for `MySqlValueGenerationStrategy.IdentityColumn` with the annotation:
[...]

PR #1329, opened 2021-02-11
Implement support for value generation strategies, that don't use an enum value but its underlying value equivalent.

lauxjpn:
Allows to use the underlying value of `MySqlValueGenerationStrategy` instead of the actual enum value (`1` in the folloing sample, instead of `MySqlValueGenerationStrategy.IdentityColumn`):

```c#
protected override void Up(MigrationBuilder migrationBuilder)
{
    migrationBuilder.CreateTable(
        name: "IceCreams",
        columns: table => new
        {
            IceCreamId = table.Column<int>(nullable: false)
                .Annotation("MySql:ValueGenerationStrategy", /*MySqlValueGenerationStrategy.IdentityColumn*/ 1), // <--
            Name = table.Column<string>(nullable: true, defaultValue: "Vanilla")
        },
        constraints: table =>
        {
            table.PrimaryKey("PK_IceCreams", x => x.IceCreamId);
        });
}
```

Addresses https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1205#issuecomment-776910444

### DIFF
--- a/src/EFCore.MySql/Internal/MySqlValueGenerationStrategyCompatibility.cs
+++ b/src/EFCore.MySql/Internal/MySqlValueGenerationStrategyCompatibility.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
 using Pomelo.EntityFrameworkCore.MySql.Metadata.Internal;
@@ -9,7 +10,10 @@ namespace Pomelo.EntityFrameworkCore.MySql.Internal
     {
         public static MySqlValueGenerationStrategy? GetValueGenerationStrategy(IAnnotation[] annotations)
         {
-            var valueGenerationStrategy = annotations.FirstOrDefault(a => a.Name == MySqlAnnotationNames.ValueGenerationStrategy)?.Value as MySqlValueGenerationStrategy?;
+            // Allow users to use the underlying type value instead of the enum itself.
+            // Workaround for: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1205
+            var valueGenerationStrategy = GetSafeEnumValue<MySqlValueGenerationStrategy>(
+                annotations.FirstOrDefault(a => a.Name == MySqlAnnotationNames.ValueGenerationStrategy)?.Value);
 
             if (!valueGenerationStrategy.HasValue ||
                 valueGenerationStrategy == MySqlValueGenerationStrategy.None)
@@ -29,5 +33,11 @@ namespace Pomelo.EntityFrameworkCore.MySql.Internal
 
             return valueGenerationStrategy;
         }
+
+        private static T? GetSafeEnumValue<T>(object value) where T : struct
+            => value != null &&
+               Enum.IsDefined(typeof(T), value)
+                ? (T?)(T)Enum.ToObject(typeof(T), value)
+                : null;
     }
 }
--- a/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.MySql.cs
+++ b/test/EFCore.MySql.FunctionalTests/MigrationSqlGeneratorMySqlTest.MySql.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata;
 using Microsoft.EntityFrameworkCore.Migrations.Operations;
 using NetTopologySuite.Geometries;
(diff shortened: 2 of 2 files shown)


JeffreySu:
@lauxjpn There is still an error when I just change the enum parameter:
https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1205#issuecomment-860732692

JeffreySu:
@lauxjpn I found when I change `MySqlValueGenerationStrategy.IdentityColumn` to `1`, the CREATE TABLE code is not the same:

Use  `MySqlValueGenerationStrategy.IdentityColumn` :
```
`Id` int NOT NULL AUTO_INCREMENT,
```

Use  `1` :
```
`Id` int NOT NULL,
```

The `AUTO_INCREMENT` config keyword is missing.

lauxjpn:
@JeffreySu I am not able to reproduce this issue.

I used the following code with Pomelo `3.2.5`:

<details>
<summary>Program.cs</summary>

```c#
using System.Diagnostics;
using System.Linq;
using Microsoft.Extensions.Logging;
using Pomelo.EntityFrameworkCore.MySql.Infrastructure;
using Pomelo.EntityFrameworkCore.MySql.Storage;

namespace IssueConsoleTemplate
{
    public class IceCream
    {
        public int IceCreamId { get; set; }
        public string Name { get; set; }
    }
    
    public class Context : DbContext
    {
        public DbSet<IceCream> IceCreams { get; set; }

        protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
        {
            if (!optionsBuilder.IsConfigured)
            {
                var connectionString = "server=127.0.0.1;port=3306;user=root;password=;database=Issue1329";
                var serverVersion = ServerVersion.AutoDetect(connectionString);
                
                //optionsBuilder.UseMySql(connectionString, serverVersion) // for EF Core 5.0+
                optionsBuilder.UseMySql(connectionString, optionsBuilder => optionsBuilder.ServerVersion(serverVersion).CharSetBehavior(CharSetBehavior.NeverAppend)) // for EF Core 3.1
                    .UseLoggerFactory(
                        LoggerFactory.Create(
                            configure => configure
                                .AddConsole()
                                .AddFilter(level => level >= LogLevel.Information)))
                    .EnableSensitiveDataLogging()
                    .EnableDetailedErrors();
            }
        }
    }

    internal static class Program
    {
        private static void Main()
        {
            using (var context = new Context())
            {
                context.Add(new IceCream {Name = "Vanilla"});
                context.SaveChanges();
            }

            using (var context = new Context())
            {
                var iceCream = context.IceCreams.First();
                
                Trace.Assert(iceCream.IceCreamId == 1);
            }
        }
    }
}
```
</details>

I then created a migration:

```
dotnet ef migrations add Initial
```

I changed its `Up()` method:

```c#
protected override void Up(MigrationBuilder migrationBuilder)
{
    migrationBuilder.CreateTable(
        name: "IceCreams",
        columns: table => new
        {
            IceCreamId = table.Column<int>(nullable: false)
                .Annotation("MySql:ValueGenerationStrategy", /* MySqlValueGenerationStrategy.IdentityColumn */ 1),
            Name = table.Column<string>(nullable: true)
        },
        constraints: table =>
        {
            table.PrimaryKey("PK_IceCreams", x => x.IceCreamId);
        });
}
```

And finally generated a script:

```
dotnet ef migrations script
```

The generated SQL script looks like the following:

```sql
CREATE TABLE IF NOT EXISTS `__EFMigrationsHistory` (
    `MigrationId` varchar(95) NOT NULL,
    `ProductVersion` varchar(32) NOT NULL,
    CONSTRAINT `PK___EFMigrationsHistory` PRIMARY KEY (`MigrationId`)
);

CREATE TABLE `IceCreams` (
    `IceCreamId` int NOT NULL AUTO_INCREMENT,
    `Name` longtext NULL,
    CONSTRAINT `PK_IceCreams` PRIMARY KEY (`IceCreamId`)
);

INSERT INTO `__EFMigrationsHistory` (`MigrationId`, `ProductVersion`)
VALUES ('20210614221810_Initial', '3.1.15');
```

As you can see, the `IceCreamId` column of the `IceCreams` table contains the `AUTO_INCREMENT` flag.

So if this is not happening on your end, we would need more of the original code, so we know what is going on.
As said earlier, I am assuming you are using Pomelo `3.2.5`.

JeffreySu:
@lauxjpn Yes, like what you said, I can reproduce the right result now, and also right in my code without any change. I don't know why 🤣 

There is an other question:
CSharpHelper seems be supported by EF Core 1.0/1.1: https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.migrations.design.csharphelper?view=efcore-1.1#applies-to

How could I use it in EF Core 5.0?

BTW I use Pomelo.EntityFrameworkCore.MySql v5.0 .

PR #1332, opened 2021-02-21
Add support for all possible `String.Concat()` variants

### LINKED ISSUE #1331: String.Concat Translation
### Steps to reproduce
Currently it's possible to concatenate strings using the method below.

```C#
var posts = _ctx.Posts.Select(p => new { 
    Url = "https://www.example.com/Post/" + p.PostId.ToString()
});
```

At the moment using [String.Concat](https://docs.microsoft.com/en-us/dotnet/api/system.string.concat) in C# doesn't translate to [CONCAT()](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_concat) in MySQL.

```C#
var posts = _ctx.Posts.Select(p => new { 
    Url = String.Concat("https://www.example.com/Post/", p.PostId)
});
```

### The issue
It would be good to translate the following overloads, see https://github.com/dotnet/efcore/issues/23859.

```
Concat(String, String)
Concat(String, String, String)
Concat(String, String, String, String)
```

lauxjpn:
Implements support for all `String.Concat()` variants, except the `ReadOnlySpan<Char>` ones.
Fixes #1331

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,7 +12,7 @@
     <PackageIconUrl>https://avatars3.githubusercontent.com/u/19828814</PackageIconUrl>
     <PackageIcon>icon.png</PackageIcon>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>latest</LangVersion>
     <DebugType>portable</DebugType>
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
--- a/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlSqlTranslatingExpressionVisitor.cs
+++ b/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlSqlTranslatingExpressionVisitor.cs
@@ -5,11 +5,12 @@ using System;
 using System.Diagnostics;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Reflection;
 using JetBrains.Annotations;
+using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Query.SqlExpressions;
 using Microsoft.EntityFrameworkCore.Storage;
-using Microsoft.EntityFrameworkCore.Utilities;
 using Pomelo.EntityFrameworkCore.MySql.Query.Expressions.Internal;
 using Pomelo.EntityFrameworkCore.MySql.Query.ExpressionTranslators.Internal;
 using Pomelo.EntityFrameworkCore.MySql.Query.Internal;
@@ -21,6 +22,12 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
     {
         private readonly IMySqlJsonPocoTranslator _jsonPocoTranslator;
         private readonly MySqlSqlExpressionFactory _sqlExpressionFactory;
+
+        protected static readonly MethodInfo[] NewArrayExpressionSupportMethodInfos = Array.Empty<MethodInfo>()
+            .Concat(typeof(MySqlDbFunctionsExtensions).GetRuntimeMethods().Where(m => m.Name == nameof(MySqlDbFunctionsExtensions.Match)))
+            .Concat(typeof(string).GetRuntimeMethods().Where(m => m.Name == nameof(string.Concat)))
+            .Where(m => m.GetParameters().Any(p => p.ParameterType.IsArray))
+            .ToArray();
 
         public MySqlSqlTranslatingExpressionVisitor(
             RelationalSqlTranslatingExpressionVisitorDependencies dependencies,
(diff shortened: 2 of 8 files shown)


mguinness:
@roji Would you be able to use any of this code for the [CONCAT()](https://w3resource.com/PostgreSQL/concat-function.php) function in npgsql?  Come to think of it, would this be better placed in [EFCore.Relational](https://github.com/dotnet/efcore/tree/main/src/EFCore.Relational) as most RDBMS have this functionality?

roji:
@mguinness @lauxjpn EF Core already has translations for string.Concat [here](https://github.com/dotnet/efcore/blob/main/src/EFCore.Relational/Query/Internal/StringMethodTranslator.cs#L76), which the Pomelo provider should be getting "for free" as a relational provider.

Note https://github.com/dotnet/efcore/issues/23859 which is about adding support for more overloads (with more than two parameters) which recently went in - is this about bringing that before 6.0 is released? Otherwise, maybe the provider is somehow causing the relational translations to not work?

mguinness:
Thanks Shay, is there any documentation about which translations are implemented upstream for relational providers?

roji:
The best docs are the source :wink: 

You can look at EF Core's [RelationalMethodCallTranslatorProvider](https://github.com/dotnet/efcore/blob/main/src/EFCore.Relational/Query/RelationalMethodCallTranslatorProvider.cs#L42) and [RelationalMemberTranslatorProvider](https://github.com/dotnet/efcore/blob/main/src/EFCore.Relational/Query/RelationalMemberTranslatorProvider.cs#L31), these set up all the translators which all relational providers get (unless overridden).

mguinness:
That makes it difficult for anyone to get a list of translations, not just upstream but also for each provider.  It makes is very hard to compare the translations between providers.  I can that being an issue for users that switch to a new provider only to find out that a method that was previously translated is no longer working.

I wonder if there would be a way using reflection to dump all classes that implement `IMethodCallTranslator` so people could discover these more easily?

lauxjpn:
@roji I basically implemented the `String.Concat()` support, because @mguinness was asking for it. So I assume, that he needs extended support (more overloads) for that in 5.0 (since he referenced the code for EF Core 6, that already implemented it on the relational level). So I used the same test names that have been used for 6, so that they will collide and our implementations can be removed once 6 is supported by us.

As for the overload support, I went the way to implement everything that could be implemented, because the time spend on the additional methods wasn't that much.
That being said, I would only consider method signatures containing the type `string` in them as crucial and assume that the others are rarely used. I do consider supporting the `params string[]` signature as important however, because if `string.Concat()` works with 4 arguments, it would be unexpected to not work with 5.

mguinness:
Thanks Laurents, I appreciate that you took the time to implement for 5.0 rather than waiting for 6.0 support.  I think that `params string[]` overload should also be supported upstream in 6.0 unless there are mitigating factors.

I went off on a tangent about a way for documenting translations as it's hard to know what each provider supports.  Often I find myself converting SQL to LINQ and it's impossible to do that without looking at the source code of the provider (and upstream) or trial and error.  There must be a better way for the average user, but I'll leave it at that for now.

roji:
@lauxjpn all clear and no problem - there's nothing problematic about bringing in those translations for 5.0 instead of waiting for 6.0 :)

@mguinness I think I misunderstood your earlier question... As a user of an EF Core provider, it's indeed reasonable to find what translations are supported. We improved the docs on this quite a bit (e.g. [for SQL Server](https://docs.microsoft.com/en-us/ef/core/providers/sql-server/functions#string-functions), [for PostgreSQL](http://www.npgsql.org/efcore/mapping/translations.html)). My link to the source code above was specifically to show what translations are implemented at the relational level, and therefore are common to all relational providers by default.

mguinness:
Thanks Shay, we copied the format from npgsql for our own [wiki page](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/wiki/Translations) - I wasn't aware of the SQL Server page.  I wonder if it would be useful to indicate upstream translations as opposed to provider translations?  How are you tracking changes for documentation?

lauxjpn:
> Often I find myself converting SQL to LINQ and it's impossible to do that without looking at the source code of the provider (and upstream) or trial and error. There must be a better way for the average user [...]

@mguinness I agree. This is probably the biggest pain when either switching providers, or just starting out with EF Core.

@roji How about a relational code analyzer. Either providers could implement their support on top of it, or they could support it by exposing an interface (or by implementing a method in their method call/member translators or something like that). The analyzer could warn users then, if they are using unsupported constructs in their queries. It might not be able to find _all_ issues (e.g. in dynamically constructed queries), but is probably good enough for most common cases.

roji:
@lauxjpn some sort of static analysis solution has been proposed many times (especially around the time we released EF Core 3.0, where client evaluation was disabled).

We'd love for something like this to exist, but at the end of the day, the feasibility of it is pretty problematic. This basically requires an analyzer to somehow statically go through the entire EF Core query pipeline - requiring a fully built EF model - and (efficiently!) translating your LINQ queries as users type...

Of course, we could do something much more simple and heuristic, e.g. specifically looking at method calls inside the expression tree  - without compiling the entire thing - and checking if they're on some sort of allow list. I'm also not convinced that's likely to be successful (or cover enough cases, or be easy to sync with the runtime reality as it changes).

But if I'm over-complicating this or you have other ideas, we can definitely discuss. Up to now we've felt that [quality docs](https://docs.microsoft.com/en-us/ef/core/providers/sql-server/functions) should be sufficient, in light of the huge challenge that static analysis would be

/cc @ajcvickers

roji:
> Thanks Shay, we copied the format from npgsql for our own wiki page

Great!

> I wonder if it would be useful to indicate upstream translations as opposed to provider translations?

It definitely would and we should do it... There was some discussion on whether the upstream translations should be in some single separate place, but from a user perspective (and given the relatively low number of upstream translations) I think we should just integrate them into each provider's docs.

> How are you tracking changes for documentation?

Regular issues on https://github.com/dotnet/EntityFramework.Docs

lauxjpn:
> Of course, we could do something much more simple and heuristic, e.g. specifically looking at method calls inside the expression tree - without compiling the entire thing - and checking if they're on some sort of allow list. I'm also not convinced that's likely to be successful (or cover enough cases, or be easy to sync with the runtime reality as it changes).

@roji Yeah, that is more in line of what I was thinking about. Basically, just analyze good query candidates (e.g. queries that access a `DbContext` or `DbSet` or use other popular methods). Then do some basic analysis on those queries (leaning more on the more conservative side) à la "does this call a method, in a query context that would probably need translation, that the referenced provider(s) does not explicitly support?". Though I am sure I am oversimplifying here and it would be much more complex to get something usable out of it.

roji:
> Though I am sure I am oversimplifying here and it would be much more complex to get something usable out of it.

Yeah, that's the general feeling we've had when discussing this. Though we are considering some static analysis of queries (https://github.com/dotnet/efcore/issues/21663), and I have some crazy ideas for the future...

PR #1354, opened 2021-03-20
Document that EF Core 6.0 is not supported yet

bgrainger:
I've seen a number of SO questions where someone is trying to use EF Core 6.0 Preview with Pomelo 5.0 alpha. The two libraries are incompatible, so this doesn't work.

I thought it might be helpful to explicitly document that EF Core 6.0 isn't supported in the version compatibility chart so people can find it themselves, or to refer to when answering SO questions or replying to GitHub issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The following versions of MySqlConnector, EF Core and .NET Standard are compatib
 
 Release | Branch | MySqlConnector | EF Core | .NET Standard | .NET (Core) | .NET Framework
 --- | --- | --- | --- | --- | --- | ---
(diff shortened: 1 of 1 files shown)


lauxjpn:
Is this necessary? The table already specifies, that `5.0.0-alpha.2` is compatible with EF Core `5.0.x`.

An alternative could be to constrain the Nuget packages to 5.0.0 >= EF Core < 6.0.0, since providers are always only compatible with the major version of EF Core they have been intended for.

bgrainger:
> Is this necessary?

Well, I wouldn't have thought so, but I just observed SO users apparently needing the clarification.

Constraining the NuGet version sounds like a better idea: make it explicit in the code/dependencies.

PR #1364, opened 2021-03-30
Fix #1352 InvalidOperationException when scaffolding without including referenced table

### LINKED ISSUE #1352: issue with scaffolding tables without including Foreign tables
Hi .. i was trying to scaffold-dbcontext 

first i added a simple table (no relations with other tables and no constraints ) and everything worked fine .. but when i tried the same with another table that has foreign key constraints with many tables but those tables  were not included in the proces i got an error 

the solution was to include all my DB tables (which is not logic here and makes the project size bigger)

Further technical details

Database engine: MySQL

Visual Studio version: Visual Studio 2019 16.9.2

.Net Core Version : .Net 5

Microsoft.EntityFrameworkCore Version="5.0.4"
Microsoft.EntityFrameworkCore.SqlServer Version="5.0.4"
Microsoft.EntityFrameworkCore.SqlServer.Design Version="1.1.6"
Microsoft.EntityFrameworkCore.Tools Version="5.0.4"

Pomelo.EntityFrameworkCore.MySql Version "5.0.0-alpha.2"

V1nzi:
When scaffolding a table t1, that references another table t2 while explicitely only including t1, an InvalidOperation was thrown by the tables.Single(...) in MySqlDatabaseModelFactory in GetConstraints.
Replaced it with a SingleOrDefault, so the missing referenced table no longer causes an exception and the model is instead just created without it.

### DIFF
--- a/src/EFCore.MySql/Scaffolding/Internal/MySqlDatabaseModelFactory.cs
+++ b/src/EFCore.MySql/Scaffolding/Internal/MySqlDatabaseModelFactory.cs
@@ -653,7 +653,7 @@ ORDER BY
                                 // different casing than the actual table name. (#1017)
                                 // In the unlikely event that there are multiple tables with the same spelling, differing only in casing,
                                 // we can't be certain which is the right match, so rather fail to be safe.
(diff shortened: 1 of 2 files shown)


lauxjpn (approved):
Looks all good to me.
Thank you for your contribution!

PR #1366, opened 2021-04-02
Fix MySQL json datatype charset and collation

### LINKED ISSUE #1365: CharSet should be ignored when ColumnType is JSON
### Steps to reproduce
```csharp
class Model
{
        [Column("property1", TypeName = "json")]
        public string Property1 { get; set; }
}

await dbContext.Database.EnsureCreatedAsync();
```

### The issue
CharSet should be ignored when ColumnType is JSON, otherwise it will be an invalid statement.

```
Exception message:
fail: Microsoft.EntityFrameworkCore.Database.Command[20102]
      Failed executing DbCommand (0ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
      CREATE TABLE `Model` (
          property1 json CHARACTER SET utf8 NULL;
      );
Failed executing DbCommand (0ms) [Parameters=[], CommandType='Text', CommandTimeout='30']
MySqlConnector.MySqlException (0x80004005): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CHARACTER SET utf8 NULL' at line 2
```

### Further technical details

MySQL version: 5.7 and 8.0
Operating system: Windows 10
Pomelo.EntityFrameworkCore.MySql version: 5.0.0-alpha.2
Microsoft.AspNetCore.App version: 5.0.4

xfoxfu:
Fix #1365. Disables `CHARACTER SET` and `COLLATE` clause appending on MySQL `json` columns.

### DIFF
--- a/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
+++ b/src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
@@ -1188,6 +1188,11 @@ DELIMITER ;";
 
         private static string GetColumnTypeWithCharSetAndCollation(ColumnOperation operation, string columnType)
         {
+            if (columnType.IndexOf("json", StringComparison.OrdinalIgnoreCase) >= 0)
+            {
+                return columnType;
+            }
+
             var charSet = operation[MySqlAnnotationNames.CharSet];
             if (charSet != null)
             {


src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
```diff
@@ -1188,6 +1188,11 @@ protected override string GetColumnType(string schema, string table, string name
 
         private static string GetColumnTypeWithCharSetAndCollation(ColumnOperation operation, string columnType)
         {
+            if (columnType.IndexOf("json", StringComparison.OrdinalIgnoreCase) >= 0)
```
lauxjpn, line 1191: Please add a support property to `ServerVersionSupport` (e.g. `JsonWithCharsetAndCollation`), override it it in `MariaDbServerVersion.MariaDbServerVersionSupport` and use it in your check here to verify whether charsets and collations are supported for JSON columns or not.

lauxjpn (changes requested):

src/EFCore.MySql/Migrations/MySqlMigrationsSqlGenerator.cs
lauxjpn: Also, add a test that executes the different behavior between MySQL and MariaDB.

lauxjpn: Actually, my requests are not needed anymore.

_(From https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1365#issuecomment-824461107:)_

> Actually, it turns out that MariaDB seems to force a charset of `utf8mb4` and a collation of `utf8mb4_bin` as well, when using the `json` type (alias for `longtext`), but doesn't if just `longtext` is used.
> So there doesn't need to be a distinction between MySQL and MariaDB here, as both will implicitly use the charset of `utf8mb4` and the `utf8mb4_bin` collation.
> 
> However, since `json` is just an alias for `longtext` for MariaDB, it returns the type `longtext` from `information_schema.columns` when we scaffold it. Because Pomelo will only automatically enable the JSON support for properties with the column type `json`, users would currently need to manually change the as `longtext` scaffolded columns to `json`.
> 
> However, since `10.4.3`, MariaDB will automatically add a `JSON_VALID()` check constraint to `json` columns (see [JSON Data Type](https://mariadb.com/kb/en/json-data-type/) and [MDEV-13916: Enforce check constraint on JSON type](https://jira.mariadb.org/browse/MDEV-13916)), so we can use that to identify a `longtext` column with a charset of `utf8mb4` and a collation of `utf8mb4_bin` as `json`.

lauxjpn (approved):

lauxjpn:
@xfoxfu Thank you for your contribution!

PR #1380, opened 2021-04-12
feat: make MySqlSqlNullabilityProcessor in MySqlParameterBasedSqlProcessor transient

Rynaret:
I am using Pomelo.EntityFrameworkCore.MySql 5.0.0-alpha.2 version.
I have found a problem when code like this

```csharp
async Task GetUsersAsync(List<int> userIds)
{
    using (MySqlDbContext dbContext = _dbContextFactory.CreateDbContext())
    {
        return await _dbContext
            .Users
            .Where(x => userIds.contains(x.Id))
            .Select(x => new { x.Id })
            .ToListAsync(cancellationToken);
    }
}
```

returns sql script from another thread

```sql
GetUsersAsync(new List<int> { 110 }) -> SELECT `v`.`id` FROM `Users` AS `v` WHERE (`v`.`id` = 100)
GetUsersAsync(new List<int> { 100 }) -> SELECT `v`.`id` FROM `Users` AS `v` WHERE (`v`.`id` = 100)
```

<details>
<summary>Actual code that I used for testing</summary>
<p>

```c#
while (true)
{
    System.Collections.Generic.List<Task> tasks = new System.Collections.Generic.List<Task>();
    for (int i = 0; i < 3; i++)
    {
        var task1 = Task.Run(async () =>
        {
            for (int i = 0; i < 40; i++)
            {
                await repo.GetUsersAsync(new List<int> { 100 }, default);
            }
        });
        var task2 = Task.Run(async () =>
        {
            for (int i = 0; i < 40; i++)
            {
                await repo.GetUsersAsync(new List<int> { 101 }, default);
            }
        });
        var task3 = Task.Run(async () =>
        {
            for (int i = 0; i < 40; i++)
            {
                await repo.GetUsersAsync(new List<int> { 110 }, default);
            }
        });
        var task4 = Task.Run(async () =>
        {
            for (int i = 0; i < 40; i++)
            {
                await repo.GetUsersAsync(new List<int> { 110 }, default);
            }
        });

        tasks.Add(task1);
        tasks.Add(task2);
        tasks.Add(task3);
        tasks.Add(task4);
    }

    await Task.WhenAll(tasks);

    await Task.Delay(5000);
}
```

</p>
</details>

I tried to debug and found that ln:79 `canCache` was set with value from another thread with `true` value set in ln:74
image

In this pull request I applied similar logic from here https://github.com/dotnet/efcore/blob/v5.0.0/src/EFCore.Relational/Query/RelationalParameterBasedSqlProcessor.cs#L89. After changes the problem has gone.

### DIFF
--- a/src/EFCore.MySql/Query/Internal/MySqlParameterBasedSqlProcessor.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlParameterBasedSqlProcessor.cs
@@ -14,7 +14,6 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.Internal
     public class MySqlParameterBasedSqlProcessor : RelationalParameterBasedSqlProcessor
     {
         private readonly IMySqlOptions _options;
-        private readonly SqlNullabilityProcessor _sqlNullabilityProcessor;
 
         public MySqlParameterBasedSqlProcessor(
             [NotNull] RelationalParameterBasedSqlProcessorDependencies dependencies,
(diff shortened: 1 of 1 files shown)


Rynaret:
I will appreciate if you will publish patch version to NuGet.

lauxjpn (approved):
@Rynaret Thanks for the PR!

While officially, EF Core is single threaded (not thread-safe) and does not support multiple parallel queries to a single `DbContext`, it does work in most cases anyway, so we shouldn't break it unnecessarily if avoidable.

The PR looks good.

PR #1422, opened 2021-05-07
Rename unsuitable file name to class name within the file.

wuyuansushen:
The filename lose the "Builder" apart of the name...It wastes much time for find "MySqlDbContextOptionsBuilderExtensions" class with this bad file name...


lauxjpn (approved):
Thanks, we missed that!

PR #1424, opened 2021-05-13
Syntax correction in Readme example

WardenDrew:
Small correction to remove bonus ';' and ')'


lauxjpn (approved):
Thanks, we missed that one!

PR #1444, opened 2021-06-12
Use alias in HAVING clause, to workaround MySQL/MariaDB bug

lauxjpn:
MySQL & MariaDB return an [Error: "Unknown column 'foo' in 'having clause'" when using a function in HAVING](https://bugs.mysql.com/bug.php?id=103961).

For example, the following query will not work in MySQL & MariaDB (but in basically all other major database servers):

```sql
SELECT YEAR(`o`.`OrderDate`) AS `Year`, COUNT(*) as `Count`
FROM `Orders` AS `o`
GROUP BY YEAR(`o`.`OrderDate`)
HAVING YEAR(`o`.`OrderDate`) = 1995
```

The error message is:

```
Error Code: 1054. Unknown column 'o.OrderDate' in 'having clause'
```

While using an expression in GROUP BY and ORDER BY clauses works as expected, it does not work when using them in HAVING clauses.

---

Until this bug is fixed in both database server implementations, we now implement a workaround, that uses the SELECT clause alias instead, which is valid for MySQL/MariaDB HAVING clauses and _does_ work.

### DIFF
--- a/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlBug96947WorkaroundExpressionVisitor.cs
+++ b/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlBug96947WorkaroundExpressionVisitor.cs
@@ -82,7 +82,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
             {
                 var expression = (SqlExpression)Visit(projectionExpression.Expression);
 
-                // Parameters trigger this bug as well, because the get inlined by MySqlConnector and
+                // Parameters trigger this bug as well, because they get inlined by MySqlConnector and
                 // become constant values in the end.
                 if (expression is SqlConstantExpression ||
                     expression is SqlParameterExpression)
@@ -91,14 +91,6 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
                         expression,
                         projectionExpression.Type,
                         expression.TypeMapping);
-                }
-
-                if (expression is SqlParameterExpression parameterExpression)
-                {
-                    expression = _sqlExpressionFactory.Convert(
-                        parameterExpression,
-                        projectionExpression.Type,
-                        parameterExpression.TypeMapping);
                 }
 
                 return projectionExpression.Update(expression);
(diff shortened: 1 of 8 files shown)


lauxjpn:
@ajcvickers Is there a better/proper way to add an additional projection with an alias to a `SelectExpression` (without overriding a lot of hard to maintain methods along the way) than our (interchangeable) `VisitSelect1()` and `VisitSelect2()` methods in our `MySqlHavingExpressionVisitor` class? They are definitely a bit of a hack, but work fine.

ajcvickers:
@smitpatel Any thoughts on @lauxjpn's question?

smitpatel:
What is the query being generated after change? EF Core SelectExpression doesn't allow using alias from projection at same level as of now. (We have backlog issue somewhere to use it in ORDER BY).
There is multiple way to resolve this, I can guide further once I know what is the expected query after rewrite.

lauxjpn:
@smitpatel For the sample query above, the fixed query would be:

```sql
SELECT
    YEAR(`o`.`OrderDate`) AS `Year`,
    COUNT(*) as `Count`,
    YEAR(`o`.`OrderDate`) = 1995 as `having` /* <-- add HAVING expression with `having` alias*/
FROM `Orders` AS `o`
GROUP BY
    YEAR(`o`.`OrderDate`),
    `having` /* <-- only add `having` alias if not containing aggregate function */
HAVING `having` /* <-- replace original expression with the `having` alias */
```

So if the query contains a HAVING clause expression, we want to add it to the projections, and then replace the HAVING clause expression with the alias of the projection.
If the HAVING clause expression does not contain an aggregate function, we also need to add the alias (or the original HAVING clause expression, both work fine) to the GROUP BY clause.

smitpatel:
Given current SelectExpression doesn't allow alias on same level, if you want to generate exact same query as above then need to use hacks.

Alternatively you can also pushdown the subquery before Having if having cannot have complex operation that will automatically convert it to where.
It will essentially generate SQL like this

```
SELECT `t`.`Year`, `t`.`Count`
(
    SELECT
        YEAR(`o`.`OrderDate`) AS `Year`,
        COUNT(*) as `Count`
    FROM `Orders` AS `o`
    GROUP BY
        YEAR(`o`.`OrderDate`)
) AS `t`
WHERE `t`.`Year` = 1995

lauxjpn:
@smitpatel Thanks, we will keep using our `VisitSelect2` implementation then, as it is less unsafe to maintain than the `VisitSelect1` implementation that depends on an internal constructor signature.

Issue #1410, opened 2021-04-27
Create attributes for specifying charset / collation


mc0re:
Please create (or adopt existing) attributes for specifying charset / collation on individual `string` properties. Something like:
```csharp
public sealed class Entity
{
    [HasCharset(Charsets.Ascii)]
    [HasCollation(Collations.AsciiDefault)]
    public string MacAddress { get; set; }
}
```

Issue #1199, opened 2020-10-15
Revamp collation (and charset) support to be in line with EF Core 5


lauxjpn:
Since EF Core 5 brings its own collation support to the table, we need to do some spring cleaning on our own support that we introduced in 3.0.

Afterwards, implementing #618 and #913 can be considered.

mguinness:
For reference, see [Configuring Character Sets and Collations in EF Core](https://dev.mysql.com/doc/connector-net/en/connector-net-entityframework-core-charset.html) with MySQL Connector.

lauxjpn:
@mguinness Thanks, I wasn't even aware, that Oracle had implemented its own support in the past. We could consider supporting their old mechanisms in a compatibility package (or namespace) or something similar, to make it easier for users to migrate to Pomelo. This support would likely introduce a set of extension methods (and attributes) of the same namespace they use, that would be marked as obsolete, but would forward their calls to ne new EF Core 5 implementations.

PR #1451, opened 2021-06-17
Add Pomelo specific CharSet data annotation attribute

lauxjpn:
Allows to use a Pomelo specific data annotation attribute named `MySqlCharSet` to be applied to entity classes and/or their properties/fields:

```c#
[MySqlCharSet("utf8mb4")]
protected class IceCreams
{
    public int IceCreamId { get; set; }

    public string InternationalName { get; set; }

    [MySqlCharSet("latin1")]
    public string WesternName { get; set; }
}
```

Character set delegation is handled the same way it is using the `HasCharSet()` fluent API call.

Addresses #1410

### DIFF
--- a/src/EFCore.MySql/DataAnnotations/CharSetAttribute.cs
+++ b/src/EFCore.MySql/DataAnnotations/CharSetAttribute.cs
@@ -1,0 +1,56 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System;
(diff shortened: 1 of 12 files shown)


mguinness:
Thanks Laurents, a couple of related issues at https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1199 and https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1410 for reference.

lauxjpn:
Also implements a temporary fix for https://github.com/dotnet/efcore/issues/25127, until it is fixed and merged upstream.

Issue #1484, opened 2021-08-02
Error assigning primary key


twsl:
### Steps to reproduce
I am trying to scaffold an old database with the following command: `dotnet ef dbcontext scaffold "Server=192.168.1.11;Database=<db>;User=root;Password=<pw>;" "Pomelo.EntityFrameworkCore.MySql" -v -f --prefix-output`

### The issue
Scaffolding doesn't work.

If you are seeing an exception, include the full exceptions details (message and stack trace).

```
verbose: No explicit ServerVersion was set.
verbose: ServerVersion '5.1.36-mysql' was automatically detected.
info:    Using ServerVersion '5.1.36-mysql'.
error:   Error assigning primary key for tbl1.
error:   Error assigning primary key for tbl2.
error:   Error assigning primary key for tbl3.
error:   Error assigning index for tbl1.
error:   Error assigning index for tbl1.
error:   Error assigning index for tbl2.
error:   Error assigning index for tbl3.
warn:    Could not scaffold the foreign key 'tbl1(A4Id)'. A key for 'A4Id' was not found in the principal entity type 'Tbl4'.
warn:    Could not scaffold the foreign key 'tbl2(A1Id)'. A key for 'A1Id' was not found in the principal entity type 'Tbl1'.
warn:    Could not scaffold the foreign key 'tbl3(T2Id)'. A key for 'T2Id' was not found in the principal entity type 'Tbl2'.
```
I modified the property and table names for privacy reasons.

### Further technical details

MySQL version: 5.1.36
Operating system: Windows
Pomelo.EntityFrameworkCore.MySql version: 5.0.0
Microsoft.AspNetCore.App version: 5

Other details about my project setup: The weird thing is, that I do have a second database running `5.5.57-MariaDB` where the a database with a different version of the same structure can be found and in phpMyAdmin the configuration looks identical. MariaDb works, MySql doesn't. According to phpMyAdmin all keys are set corretly in both dbs.

twsl:
Simply trying to copy the configuration from the other database results in:
`MySqlConnector.MySqlException (0x80004005): Incorrect key file for table '/tmp/#sql_28ce_0.MYI'; try to repair it`.
Tablecheck in phpMyAdmin says everything is fine.

mguinness:
[...]

PR #1486, opened 2021-08-06
Proposed fix for #1484

twsl:
Fix for #1484 proposed by @mguinness for an weird quirk in an old, unsupported version of MySql.

### DIFF
--- a/src/EFCore.MySql/Infrastructure/MariaDbServerVersion.cs
+++ b/src/EFCore.MySql/Infrastructure/MariaDbServerVersion.cs
@@ -82,7 +82,6 @@ namespace Microsoft.EntityFrameworkCore
             public override bool InformationSchemaCheckConstraintsTable => ServerVersion.Version >= new Version(10, 3, 10) ||
                                                                            ServerVersion.Version.Major == 10 && ServerVersion.Version.Minor == 2 && ServerVersion.Version.Build >= 22;  // MySQL is missing the explicit TABLE_NAME column that MariaDB supports, so always join the TABLE_CONSTRAINTS table when accessing CHECK_CONSTRAINTS for any database server that supports CHECK_CONSTRAINTS.
             public override bool IdentifyJsonColumsByCheckConstraints => true;
-            public override bool MySqlWL2649Workaround => false;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


mguinness:
Don't you have to make the same changes in `GetIndexes()`?

You'll have to wait for lauxjpn to take a look at the build errors which seem to be occuring in `MySqlJsonMicrosoftTypeMapping`.

lauxjpn:
> I created a draft PR, but the tests fail and I cannot see the azure pipelines error.

You should be able to see them. They are public:

image

image

image

image

The pipeline fails because of some changes in the latest .NET preview, that leads to some ambiguities. This is unrelated to this issue here.
I will upload a fix for the pipeline issue, that then limits the .NET version to a specific one.

lauxjpn (dismissed):
I think it is safe to just use `CAST()` for all versions.
Please add a test or two that would fail without your code fix.

twsl:
Thanks for the extensive help, not sure why I wasn't able to see them previously; now I don't see the azure error I used to see before. I'll try to come up with tests asap.

lauxjpn:
> Yet I fail to understand, why the pipeline fails, the error messages are not really helpful.

The pipeline is currently configured to use the latest preview of .NET 6. In the latest iteration, they have made some changes to the `Lazy<>` class that lead to some ambiguous calls where .NET cannot say with certainty anymore, what method (in this case constructor) was intended to be called.

I hit the same issue locally a couple of weeks ago when trying to compile Pomelo against the latest .NET preview.

twsl:
> I think it is safe to just use `CAST()` for all versions.
> Please add a test or two that would fail without your code fix.

That would make the Version.Supports property useless and it can be safely removed, right?
And I am unsure how to structure a test and what to mock with the cast fix, as the specific MySql version simply returns a blob instead of a char inside the query. Can you provide some directions @lauxjpn ? Are tests even required?

lauxjpn:
> Are tests even required?

I think it is probably enough to just add a test to `MySqlDatabaseModelFactoryTest`'s primary key facets and index facets region respectively, to demonstrate that the `CAST()` change does not break the execution of the query.
Just verify locally on your end that the test case fails with MySQL 5.1 without the fix, because we don't test MySQL 5.1 in CI.

> That would make the Version.Supports property useless and there is no need for it, right?

Yes, I don't think we need it.

---

It is possible to use Moq or manually mock it in a way, that you change the return type to `byte[]` instead of `string` for `SUB_PARTS`, but that is probably overkill as we want to keep this fix as simple as possible.

twsl:
I don't think that additional tests are required because both functions are called every time a test database is initialized within the existing tests. If they don't fail, everything should be fine.
Due to errors with c#10 language features in VS and/or the already mentioned lazy build errors I cannot properly test everything.

lauxjpn:
> Due to errors with c#10 language features in VS and/or the already mentioned lazy build errors I cannot properly test everything.

Even with those fixed, the tests will not run against MySQL 5.1, because they use features like the `utf8mb4` charset, that were not supported back then.

Just rebase your PR on top of the latest commit, which fixes the `Lazy<>` issue, so the CI succeeds.
We'll merge it then.

twsl:
@lauxjpn can I click the merge button myself? I always forget that someone once made me a team member 😄

PR #1497, opened 2021-08-30
Add translation of DateTime.AddMilliseconds

### LINKED ISSUE #1496: DateTime.AddMilliseconds method not supported
### The issue

For some reason I need to use the `DateTime.AddMilliseconds` method to remove the millisecond in the `DateTime`, and this cause an exception.

```
Translation of method 'System.DateTime.AddMilliseconds' failed.
```

And I'd like to know if this method will be translated in the future.

### Further technical details
 
Pomelo.EntityFrameworkCore.MySql version:5.0.1

iamSmallY:
Close #1496 

I use 
```mysql
date_add(@dt, interval num * 1000 microsecond);
```
to translate the
```C#
dt.AddMilliseconds(num);
```

Since the `AddMilliseconds` method will round the parameter to the nearest integer, I also add four tests for round up and round down.

### DIFF
--- a/src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq.Expressions;
+using System.Data;
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Diagnostics;
@@ -65,7 +65,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.Internal
                                     _sqlExpressionFactory.Fragment("INTERVAL"),
                                     datePart.Equals("microsecond")
                                         ? _sqlExpressionFactory.Multiply(
-                                            new SqlConstantExpression(Expression.Constant(1000), new IntTypeMapping("int")),
+                                            _sqlExpressionFactory.Constant(1000, new IntTypeMapping("int32", DbType.Int32)),
                                             _sqlExpressionFactory.Convert(arguments[0], typeof(int)))
                                         : _sqlExpressionFactory.Convert(arguments[0], typeof(int)),
                                     _sqlExpressionFactory.Fragment(datePart)
--- a/src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Collections.Generic;
-using System.Data;
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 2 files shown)


src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
```diff
@@ -59,7 +63,11 @@ public MySqlDateTimeMethodTranslator(ISqlExpressionFactory sqlExpressionFactory)
                                 new SqlExpression[]
                                 {
                                     _sqlExpressionFactory.Fragment("INTERVAL"),
-                                    _sqlExpressionFactory.Convert(arguments[0], typeof(int)),
+                                    datePart.Equals("microsecond")
+                                        ? _sqlExpressionFactory.Multiply(
+                                            new SqlConstantExpression(Expression.Constant(1000), new IntTypeMapping("int")),
```
lauxjpn, line 68: Please use `_sqlExpressionFactory.Constant(1000)` here.

lauxjpn (changes requested):
Thanks for adding the translation!
The SQL Server provider implements it as well.

src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
```diff
@@ -59,7 +63,11 @@ public MySqlDateTimeMethodTranslator(ISqlExpressionFactory sqlExpressionFactory)
                                 new SqlExpression[]
                                 {
                                     _sqlExpressionFactory.Fragment("INTERVAL"),
-                                    _sqlExpressionFactory.Convert(arguments[0], typeof(int)),
+                                    datePart.Equals("microsecond")
+                                        ? _sqlExpressionFactory.Multiply(
+                                            _sqlExpressionFactory.Constant(1000, new IntTypeMapping("int32", DbType.Int32)),
```
lauxjpn, line 68: Is the explicit type mapping really necessary? I would have assumed, that it works just fine without it.

lauxjpn (changes requested):
Also please rebase your changes on top of `master`, instead of `5.0-maint`. Thanks!

src/EFCore.MySql/Query/Internal/MySqlDateTimeMethodTranslator.cs
```diff
@@ -3,11 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Reflection;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Query.SqlExpressions;
+using Microsoft.EntityFrameworkCore.Storage;
```
lauxjpn, line 12: The two additional `using` statements should not be needed in the current code.

lauxjpn (changes requested):

lauxjpn (approved):

lauxjpn:
@iamSmallY Thank you for your contribution!

iamSmallY:
No thanks😀

Issue #1413, opened 2021-04-28
EF Core 6 support


lauxjpn:
We will use this issue to track the progress on supporting EF Core 6, including its previews.

lauxjpn:
With #1416, we have now full support for the upcoming **EF Core 6.0 Preview 4**.

The code quality should be as high as 5.0.0 GA, with all old and new tests working.
Therefore, the provider should run very stable.

We will push it to nuget when EF Core preview 4 gets released. Until then, use our [nightly builds](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql#nightly-builds) together with the ones from EF Core. For the latter, add their feed to your [NuGet.config](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/NuGet.config) file (here with the key `dotnet6`):

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="pomelo" value="https://pkgs.dev.azure.com/pomelo-efcore/Pomelo.EntityFrameworkCore.MySql/_packaging/pomelo-efcore-public/nuget/v3/index.json" />
    <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

mguinness:
Thanks for giving up your free time this past weekend to get this over the finish line.  Your hard work is very much appreciated.

Just a reminder to others that this is a community based project so if you are able to contribute either by helping with issues or with documentation that would also be appreciated.

lauxjpn:
> One area we need help with is documenting both JSON and spatial queries. See [JSON Mapping](https://www.npgsql.org/efcore/mapping/json.html?tabs=data-annotations%2Cpoco) and [Spatial Mapping](https://www.npgsql.org/efcore/mapping/nts.html) as an example.

I'll attach the latest version of the function/method translations Excel file, that I use internally, to this post:

[Translations_20201231_01.xlsx](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/files/6488730/Translations_20201231_01.xlsx)

lauxjpn:
[...]

PR #1531, opened 2021-10-05
Upgrade to EF Core 6.0.0-rc1

lauxjpn:
Addresses #1413


src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlHavingExpressionVisitor.cs
```diff
@@ -50,9 +50,18 @@ protected virtual Expression VisitSelect(SelectExpression selectExpression)
...
                         new HatSwappingColumnExpression(
-                            havingExpression,
+                            new HatSwappingColumnExpression(
+                                havingExpression,
+                                alias,
+                                havingExpression.Type,
+                                havingExpression.TypeMapping),
                             alias,
                             havingExpression.Type,
                             havingExpression.TypeMapping));
```
TheConstructor, line 67: Why not just
```cs
                    var projectionIndex = subQuery.AddToProjection(havingExpression);
                    var alias = subQuery.Projection[projectionIndex].Alias;
```

We would loose any control over the name, but no longer risk having to adopt this code every other version of EF Core

lauxjpn: If it works and is simpler or has less maintenance overhead, go ahead and open a PR!

TheConstructor: Created #1681

PR #1645, opened 2022-05-11
Add support for `string.IndexOf(string value, int startIndex) `

LeaFrock:
Fix issue #1625

PR #1652, opened 2022-05-16
Optimize `string.IndexOf(string, int)` translation

LeaFrock:
Based on [PR for efcore](https://github.com/dotnet/efcore/pull/28031) from @roji, the SQL translation can be optimized when `startIndex` is a constant. 
It's also a small enhancement for my previous PR #1645 .


lauxjpn (approved):

lauxjpn:
@LeaFrock Thank you for your contribution!

PR #1682, opened 2022-07-09
Change `Match` extension methods and introduce `IsMatch` ones

### LINKED ISSUE #1679: How to get relevance score on full text search with EF Core
Hi, with stored procedure we can get score like this

```SQL
SELECT id, MATCH (title,content,author) AGAINST ('traveling to parks') as score FROM news;
```
```Shell
+----+----------------------+
| id | score                |
+----+----------------------+
|  1 | 0.031008131802082062 |
|  2 |  0.25865283608436584 |
|  2 |  0                   |
+----+----------------------+
3 rows in set (0.00 sec)
```

And with EF Core we can get similiar result using code like this
```C#
var result = _db.news
    .Where(p => EF.Functions
    .Match(new string[]
    { p.title,p.content,p.author },
        "traveling to parks", MySqlMatchSearchMode.NaturalLanguage));

return result.ToList();
```

Is it possible to get relevance score on full text search with EF Core just like in stored procedure?

lauxjpn:
The `Match()` methods should return the relevance value of the match in the same way as the `MATCH...AGAINST` clause does in MySQL.

For convenience, we keep the previous functionality to just return `true` or `false` around, but rename those methods to `IsMatch()`, that just translate to `MATCH() ... AGAINST() > 0.0`.

Fixes #1679

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
+++ b/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
@@ -662,10 +662,54 @@ namespace Microsoft.EntityFrameworkCore
         /// <param name="_">The DbFunctions instance.</param>
         /// <param name="property">The property of entity that is to be matched.</param>
         /// <param name="pattern">The pattern against which Full Text search is performed</param>
-        /// <param name="searchMode">Mode in which search is performed</param>
+        /// <param name="searchMode">The mode to performed the search with.</param>
         /// <returns>true if there is a match.</returns>
         /// <exception cref="InvalidOperationException">Throws when query switched to client-evaluation.</exception>
-        public static bool Match(
+        public static bool IsMatch(
+            [CanBeNull] this DbFunctions _,
+            [CanBeNull] string property,
+            [CanBeNull] string pattern,
+            MySqlMatchSearchMode searchMode)
+            => throw new InvalidOperationException(CoreStrings.FunctionOnClient(nameof(IsMatch)));
+
+        /// <summary>
+        ///     <para>
+        ///         An implementation of the SQL MATCH operation for Full Text search.
+        ///     </para>
+        ///     <para>
+        ///         The semantics of the comparison will depend on the database configuration.
+        ///         In particular, it may be either case-sensitive or case-insensitive.
+        ///     </para>
+        /// </summary>
+        /// <param name="_">The DbFunctions instance.</param>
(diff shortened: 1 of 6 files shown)


TheConstructor:
Did you intentionally skip the `arguments[3] is SqlParameterExpression parameter` case for the new `.Match`-mapper?

lauxjpn:
@TheConstructor I'm not sure which line you are referring to. Just directly comment on the line (using the `Files changed` tab).

TheConstructor:
I guess you could try a `CASE WHEN` expression, and keep recommending to use a constant value.

lauxjpn:
> Just re-read the comment. I guess using index would be interesting, if used in ORDER BY, and Maria DB-support would also be great, but implementing `CASE WHEN` shouldn't be worse than no translation.

The chosen path that EF Core takes (which I think is reasonable), is to provide only performant options by default, because most users will not investigate performance issues and just expect things to work well out-of-the-box.

That provides us with the following options:

1. Only support a constant value and throw otherwise (current PR).
2. Make a partially unoptimized implementation (like using `CASE`) opt-in.
3. Support only optimized/index using `CASE` implementations and throw otherwise (i.e. for MariaDB).
4. Revert to using explicit methods (like `MatchBooleanMode()`) that kind of force users to explicitly choose a mode.

Which I see as follows:

1. is simple, but with a higher chance to throw at runtime.
2. seems too much work to support a specific corner case.
3. is reasonable, but might be confusing when switching databases.
4. is what other providers do: more redundant but kind of forces the users hand.

So in my opinion, all options except 2 are probably fine, all having their individual pros and cons.

@TheConstructor Do you think we should implement option 3 instead of option 1?

TheConstructor:
I dislike, that the exceptions are only occuring at runtime, as it is not always trivial to test all possible query variants. The user is already able to implement CASE WHEN, so maybe an explaining exception would be the best approach. 4 would make upgrading more difficult, but would remove this problem completely.

So after your explanation I would probably add a specific exception, that states that the last argument needs to be a constant for now.

Another question, that came to my mind, is, if we should give the user a way to pass a value other than 0.0 to IsMatch.

lauxjpn:
@TheConstructor Sorry, I forget this one is still open! Since we have already published a silver release, this PR will unfortunately not make it into 7.0.x anymore. :disappointed:

TheConstructor:
@lauxjpn so when to expect some movement now? 8.0 sounds like autumn?

lauxjpn:
@TheConstructor This should make it into `8.0.0-beta.2`.

PR #1687, opened 2022-07-21
Prepare release

lauxjpn:


TheConstructor:
@lauxjpn thank you for releasing 6.0.2! Just noticed the README.md still talks about 6.0.1. Surely not critical, but could be adjusted. Maybe along with a new row stating that master is in preparation of 7.x

PR #1697, opened 2022-07-27
Adding github actions for automated tests without an azure account.

renemadsen:
Adding dependabot for easy testing of dependency updates.

I made the changes to allow contributers like my self to be able to see test results, without having to do a PR, since Github actions does not require an azure account and just runs with the changes a person does.

I added the dependabot, since we are using Pomelo.EntityFrameworkCore.MySql in a lot of projects and would like to have dependencies updated faster/easier. This can be done with letting dependabot run tests for new updates and the gab of breaking changes can be minimized by doing micro updates of dependencies.


lauxjpn:
We will merge this PR into its own temporary branch for now, rework it a bit and generate new PRs for the `8.0.0-beta.2` release.

lauxjpn (approved):

lauxjpn:
@renemadsen Thank you for your contribution!

PR #1744, opened 2023-01-16
Ensure AUTOCOMMIT is set to 1 where needed

### LINKED ISSUE #1740: Changes not committed by SaveChanges for single-statement updates (probably EF7-related)
### Summary

After upgrading a project to EF7 and Pomelo.EntityFrameworkCore.MySql 7.0.0-silver.1, certain changes to the database stopped working, without any errors. Everything seemed to work, but the changes were never committed to the database. As far as we can see, this affected only changes that generated single statements (as in deleting a record from a table).

The details haven't been fully investigated, but we finally found a workaround, and I'm creating this issue in case others encounter the same problem - there's also a chance that the cause is obvious to the maintainers when they see this. I'll try to do some further testing and research over the weekend.

### Steps to reproduce

In a .NET 7.0 (Windows x64) project, we use these nuget packages (and whatever else is added automagically):

- Microsoft.EntityFrameworkCore 7.0.1
- Pomelo.EntityFrameworkCore.MySql 7.0.0-silver.1
- MySqlConnector 2.2.5

We connect to a database that runs MySQL version 5.6.21-log, using InnoDB. Yes, it's old but that's out of our control at the moment.

Configuration of our DbContext:

```
            services.AddDbContext<TestDbContext>(option =>
            {
                option.UseMySql(testDbConnectionString, new MySqlServerVersion("5.6.21-log"))
                    .EnableSensitiveDataLogging()
                    .EnableDetailedErrors();
            });
```

### The issue

Single-statement database changes (SaveChanges) *seem* to work according to all logging, but the changes never get committed.

Running this on a freshly injected context results in one affected record, but it's still in the database (note that CurrentTransaction on context.Database is null, as expected, before the second line is executed):

```
context.Remove<SomeEntity>(entity);
var count = context.SaveChanges(); // count is 1 after execution
```

No exceptions, no error messages. The log looks like this:

```
Microsoft.EntityFrameworkCore.Database.Command: Information: Executed DbCommand (17ms) [Parameters=[@p0='1234'], CommandType='Text', CommandTimeout='30']
DELETE FROM `SomeTable`
WHERE `Id` = @p0;
SELECT ROW_COUNT();
```

### Workaround

Add this in the constructor for the context:

```
Database.AutoTransactionBehavior = AutoTransactionBehavior.Always;
```

This makes the problem go away.

### Suspicion

When AutoTransactionBehaviour is the default (AutoTransactionBehaviour.WhenNeeded), autocommit (MySql "SET autocommit=0") is set to 0 somewhere, which is wrong (it doesn't turn off implicit transactions, it just turns off committing them automatically). InnoDB starts implicit transactions ALWAYS, and it can't be turned off - a commit is ALWAYS required, disabling autocommit just stops it from happening automatically.

### Further details

See my comments on #1620 - as you can see, I got some help from @roji on this.

This is the change in EF7 that this bug is most likely caused by:
https://learn.microsoft.com/en-us/ef/core/what-is-new/ef-core-7.0/whatsnew#unneeded-transactions-are-eliminated

lauxjpn:
Similar to SQL Server, we force a `SET AUTOCOMMIT = 1;` statement before statements, that are executed without explicit transaction, just in case the server or session default was previously changed to `AUTOCOMMIT = 0`.

Fixes #1740

### DIFF
--- a/src/EFCore.MySql/Update/Internal/MySqlUpdateSqlGenerator.cs
+++ b/src/EFCore.MySql/Update/Internal/MySqlUpdateSqlGenerator.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Pomelo Foundation. All rights reserved.
 // Licensed under the MIT. See LICENSE in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Globalization;
@@ -343,6 +344,9 @@ namespace Pomelo.EntityFrameworkCore.MySql.Update.Internal
 
             return isIdentityOperation;
         }
+
+        public override void PrependEnsureAutocommit(StringBuilder commandStringBuilder)
+            => commandStringBuilder.Insert(0, $"SET AUTOCOMMIT = 1{SqlGenerationHelper.StatementTerminator}{Environment.NewLine}");
 
         private static (string tableName, string schema) GetTableNameAndSchema(IColumnModification modification, IProperty property)
         {
--- a/test/EFCore.MySql.FunctionalTests/DataAnnotationMySqlTest.cs
+++ b/test/EFCore.MySql.FunctionalTests/DataAnnotationMySqlTest.cs
@@ -114,33 +114,43 @@ namespace Pomelo.EntityFrameworkCore.MySql.FunctionalTests
             base.ConcurrencyCheckAttribute_throws_if_value_in_database_changed();
 
             AssertSql(
-                @"SELECT `s`.`Unique_No`, `s`.`MaxLengthProperty`, `s`.`Name`, `s`.`RowVersion`, `s`.`AdditionalDetails_Name`, `s`.`AdditionalDetails_Value`, `s`.`Details_Name`, `s`.`Details_Value`
+                """
+SELECT `s`.`Unique_No`, `s`.`MaxLengthProperty`, `s`.`Name`, `s`.`RowVersion`, `s`.`AdditionalDetails_Name`, `s`.`AdditionalDetails_Value`, `s`.`Details_Name`, `s`.`Details_Value`
 FROM `Sample` AS `s`
 WHERE `s`.`Unique_No` = 1
-LIMIT 1",
+LIMIT 1
+""",
                 //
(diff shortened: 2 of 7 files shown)


nightcoder62:
@lauxjpn : I was the one writing the original issue, and in the exact same system I ran into the same problem again, but this time with the new ExecuteDelete method. I think this fix is not implemented there. Wrapping the work in a transaction works well, so we have no issues at the moment, but perhaps it should be implemented for any other actions (ExecuteDelete, ExecuteUpdate, ExecuteSql and so on) that may experience the same problem with that setting?

Again - I acknowledge that it's really weird and probably rare to have autocommit=0 as a default in a server, but like I wrote before, we can't change it due to some legacy database tool that is still in use here ...

roji:
@nightcoder62 currently, AutoTransactionBehavior applies only to SaveChanges, and not to ExecuteUpdate and ExecuteDelete; this is by design. As another example, when a LINQ query produces multiple SQL queries (i.e. because of [split query](https://learn.microsoft.com/en-us/ef/core/querying/single-split-queries#split-queries)), no transaction is started regardless of AutoTransactionBehavior, although such a transaction could guarantee a consistent view across the multiple queries (depending on the isolation level).

In other words, the same behavior currently happens on the SQL Server provider as well. If you think this is wrong, you can open an issue on the EF repo and we can discuss this.

nightcoder62:
@roji I suspected that this was the case. Interesting. Again, it must definitely be an edge case to have the server setting (or the SQL Server variety of it) we have, so I'm not decidedly thinking the current EF (or provider) behaviour is wrong - I could easily accept it as being "by design" and be OK with it. In either case, thanks for confirming that it is indeed an EF "problem" (if it is a problem) and not specific to the MySQL provider.

Also, I haven't studied the docs (but I will). If I think they can be improved (and feel that it's needed), I'll open on issue there.

Thanks again!

roji:
@nightcoder62 great. It would be good if you could open the issue so we can at least discuss it internally and make sure we're all aligned on this behavior.

nightcoder62:
@roji : As it's almost midnight here in Sweden (and my nick is no longer as relevant as it was 20 years ago - especially not on a Friday night after a Scotch or two), I'll let it rest until tomorrow. And possibly think it through before I post it. But yes, there'll be an issue over on EF core in a day or two. :D

roji:
Oh no urgency at all @nightcoder62 :) Just wanted to make sure this comes up in our discussions at some point, is all. Thanks!

nightcoder62:
@roji Added an issue last week, so we'll see what happens. Thank's for being there! And that, of course, goes to @lauxjpn  as well. :D

PR #1785, opened 2023-08-08
Change `Match` extension methods to return the relevance value.

yukozh:
Introduce `IsMatch` extension methods, that just return `true` or `false`.


yukozh:
Close #1785

Issue #1746, opened 2023-01-16
EF Core 8 support


lauxjpn:
We will use this issue to track the progress on supporting EF Core 8, including its previews.

tdtc-hrb:
> We will use this issue to track the progress on supporting EF Core 8, including its previews.

Support, look forward to...

DeepWorksStudios:
We would very much appreciate it if you would add support for EF Core 8 soon, as we would like to continue using maria DB.

JamesAXapien:
Any updates?

yukozh:
I'm working on upgrade Pomelo.EFCore.MySql to 8.0-preview6, there are 20 cases failing now.

yukozh:
2023-08-08

![image](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/assets/2216750/34459e54-bbfd-433e-b059-61bda5989274)

About 346 failed cases in functional tests.

Ryangr0:
Any progress here? There's lots of us that want to work with cutting edge stuff, but if you're dealing with for example MySql, it's difficult to find a workaround for making concrete implementations of repositories, basically, because everything online has relied on this third party package for years. I haven't made it work yet, so far.

DeepWorksStudios:
I agree is there any progress ?

Seabizkit:
little confused as to why there is two my understanding is cozs mysql become like private and thats not free which is the the mysql provider... and as such mysql was forked and marinadb was bore but cozs the split we now need another db provider aka this package repo. as there are probably millions of ppl moved to free version there are probably millions wanting to migrate to .net 8 and marinadb... im surprised there isnt even a beta version for support out yet.. but also as this is free it may take some time ,am i missing something.. do we know if this is being worked on and the progress to get a preview

lauxjpn:
_Prerelease_ [8.0.0-beta.1](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/releases/tag/8.0.0-beta.1) (compatible with EF Core 8.0 RC1) is now available on [nuget.org](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql/8.0.0-beta.1).

Everything that worked in 7.0.0 should still work.
The release is backed by 29,527 tests.

[...]

PR #1791, opened 2023-09-29
Upgrade to EF Core 8.0.0-rc.1

lauxjpn:
The upgrade is compatible with EF Core 8.0 RC1.

Everything compiles and all supported tests are green for all officially supported database engine versions:

- [x] MySQL 8.0
- [x] MySQL 5.7
- [x] MariaDB 11.1
- [x] MariaDB 11.0
- [x] MariaDB 10.11
- [x] MariaDB 10.10
- [x] MariaDB 10.6
- [x] MariaDB 10.5
- [x] MariaDB 10.4

The following versions have reached their end-of-live and are not actively tested by us anymore (but are expected to work anyway):

* MariaDB 10.9
* MariaDB 10.8
* MariaDB 10.7
* MariaDB 10.3

We currently have the following test statistics (here for MySQL 8):

Test suite | Tests | Succeeded | Failed | Skipped
--- | --- | --- | --- | ---
Functional | 29,527 | 29,144 | 0 | 383
Integration | 37 | 37 | 0 | 0
Custom | 51 | 51 | 0 | 0

### Primitive collections support

There are some serious bugs in the implementation of `JSON_TABLE()` in both, MySQL and MariaDB.
Because of that, we have disabled primitive collections support by default for now.
It can be enabled using the `EnablePrimitiveCollectionsSupport` option. If you do, make sure you read about possible issues below.

#### MySQL

While current versions of MySQL pretty much implement the whole spectrum that we need for propper primitive collections support, there are multiple fatal bugs that will crash the MySQL database engine (not the Pomelo provider, but the actual `mysqld` process).

The main crash is connected to using a parameter that contains a JSON array as the source of the JSON_TABLE() call (e.g. `... FROM JSON_TABLE(@p0, ...)` with `@p0 = "[1,2,3]"`). The bug kind of non-deterministic and only crashes the database engine process after a couple of those queries.
We work around this bug by inlining those constant values directly into the SQL, instead of using the parameter.

There is at least one more bug, that is deterministic and crashes the database engine process immediately every single time. This can be reproduced by running the following two tests:
* `NorthwindSelectQueryMySqlTest.Correlated_collection_after_distinct_not_containing_original_identifier(bool async)`
* `NorthwindSelectQueryMySqlTest.Correlated_collection_after_groupby_with_complex_projection_not_containing_original_identifier(bool async)`

#### MariaDB

MariaDB does not crash, but it still does not supports `LATERAL` join support (which is in many cases implicitly needed to make `JSON_TABLE()` useful).

If aggregates are used over `JSON_TABLE()` (e.g. `COUNT(*)`), then a referenced column from the outer table will always contain the value of the first row, instead of the current row, which is very much unexpected.
Referencing columns further up is not supported at all (due to missing `LATERAL` join support).
There are additional cases, where MariaDB behaves oddly.
This results in many primitive collections queries not working, that make use of `JSON_TABLE()`.

---

Addresses #1746


roji:
Ouch, sorry to hear about these database engine issues @lauxjpn... Hopefully getting everything working in the old constant mode didn't present too many difficulties (as we also do in SQL Server when the compatibility level is too old).

Any issues on the MySQL/MariaDB side tracking these?? It certainly isn't great if the engine itself crashes :sweat_smile:

lauxjpn:
> Hopefully getting everything working in the old constant mode didn't present too many difficulties (as we also do in SQL Server when the compatibility level is too old).

Yeah, that is what we are currently doing. We will cleanup some code for the next release and will probably duplicate the test class a couple of times for the different database engine version and workaround cases.

> Any issues on the MySQL/MariaDB side tracking these?? It certainly isn't great if the engine itself crashes :sweat_smile: 

I have not filed the bugs yet, but will over the next days, after I assembled the respective MREs. If Oracle isn't fixing this in a timely manner, I might do it myself to figure out appropriate workarounds along the way that we can implement.

roji:
> [...] We will cleanup some code for the next release and will probably duplicate the test class a couple of times for the different database engine version and workaround cases.

Yeah, we similarly have two implementations of PrimitiveCollectionsQueryRelationalTestBase, one for modern SQL Server and one with the compatibility flag set, where the older expand-to-constant strategy is used ([source code](https://github.com/dotnet/efcore/blob/release/8.0/test/EFCore.SqlServer.FunctionalTests/Query/PrimitiveCollectionsQueryOldSqlServerTest.cs)).

> I have not filed the bugs yet, but will over the next days, after I assembled the respective MREs. If Oracle isn't fixing this in a timely manner, I might do it myself to figure out appropriate workarounds along the way that we can implement.

Good luck! It will be interesting to follow what happens with this, and I hope MySQL/MariaDB get to a place where JSON_TABLE can be used safely by anyone...

ajcvickers:
@lauxjpn Really appreciate your hard work on this, as always.

roji:
Indeed!

gthvidsten:
@lauxjpn Late to the party here, but I'm currently looking into primitive collections and especially the `EnablePrimitiveCollectionsSupport` switch. Do you know if this has been fixed in later versions of MySQL? If not, Is there a report in their bug tracker you could link to so that I can follow their progress?

gthvidsten:
The order of the options extensions is important. If I move `EnablePrimitiveCollectionsSupport()` to be after `.MigrationsAssembly()` then the error goes away.

However, a standard EF query that produces correct results in a SQLite database, does not produce any result at all in MySQL so there might still be some issues with the implementation.

PR #1795, opened 2023-10-08
Move CI from Azure DevOps (AZDO) to GitHub Actions

lauxjpn:
We move our CI from Azure DevOps to GitHub Actions, so that anybody can run the test suit for their fork, before opening a PR.
Also, GitHub Actions has fewer limitations, is much simpler, has fewer hiccups and performs better.

We use our own [MySQL](https://www.myget.org/feed/pomelo/package/nuget/Pomelo.Chocolatey.MySQL.Server) and [MariaDB](https://www.myget.org/feed/pomelo/package/nuget/Pomelo.Chocolatey.MariaDB.Server) Chocolatey packages on Windows runners, because the official packages are broken and not up-to-date.

This allows us now to run all latest supported database engine versions on Windows during CI as well.


roji:
Congrats on this @lauxjpn!

BTW I've been toying with the idea of using [testcontainers](https://testcontainers.com) to spin up the database programmatically for the tests, rather than relying on external installation via e.g. Chocolatey. Aside from making the setup the same across Linux/Windows (it manages the docker spin-up for you), this also has the advantage of making the tests pass on developer machines without having to do any setup whatsoever.

Just in case this is interesting to you...

lauxjpn:
@roji That would make things easier for developers who want to contribute. Having to do any setup at all after cloning a repo is definitely a barrier. Installing docker is probably easier than installing and configuring a database (or a second instance of one).

With tests now also running on forks for us, we basically moved the setup away from databases to enabling Actions on GitHub (which could be a discoverability issue). And of course you have to commit and push your changes every single time to trigger the tests.

Since we currently support nearly 10 different database release trains, I naturally run tests locally against docker hosted databases but admittedly only for containers based on Linux images, because it's too much effort (and can be a bit brittle) to switch between Linux and Windows containers. (Also, the Windows containers are huge in size.)

So let me know if the Testcontainers scenario works out and simplifies things, when you get to it in the future.

(Regarding Chocolatey, there where restrictions that we hit on AZDO Pipelines some years ago when using Windows based docker database images due to their size. Though it might be worth trying again with GitHub Actions at some point.)

Issue #1793, opened 2023-10-06
Enable RenameIndex for MariaDbServerVersion >= 10.5.2


JustusGreiberORGADATA:
### The issue

MariaDB supports RenameIndex for version 10.5.2 or higher:

https://jira.mariadb.org/browse/MDEV-7318

But it is not marked like that in MariaDbServerVersion:

https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/master/src/EFCore.MySql/Infrastructure/MariaDbServerVersion.cs#L54-L56

PR #1802, opened 2023-10-17
Add RenameIndex support for MariaDB versions >= 10.5.2

JustusGreiberORGADATA:
Rename index support was added by the following MariaDB issue: https://jira.mariadb.org/browse/MDEV-7318

As far as I can see it is the exact same syntax, that MySQL supports. This syntax is already supported by this library and tests exist as well.

Changing this makes migrations for MariaDB easier, because you can't run into the "Cannot drop index needed in a foreign key constraint" problem.

Fixes: #1793


lauxjpn:
@JustusGreiberORGADATA Thank you for your contribution!

PR #1813, opened 2023-11-21
Pass loggerFactory into MySqlConnection

trejjam:
Hi, I am not completely sure this is how it should be implemented - speaking about passing ILoggerFactory. But I would love to have a native way to attach the logger pipeline into mysqlConnector.

I would be glad to receive any guidance on how to make this or similar things accepted

### DIFF
--- a/src/EFCore.MySql/Storage/Internal/MySqlRelationalConnection.cs
+++ b/src/EFCore.MySql/Storage/Internal/MySqlRelationalConnection.cs
@@ -7,9 +7,11 @@ using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Transactions;
+using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Storage;
+using Microsoft.Extensions.Logging;
 using MySqlConnector;
 using Pomelo.EntityFrameworkCore.MySql.Infrastructure;
 using Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal;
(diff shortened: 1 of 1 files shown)


roji:
@trejjam this is the wrong approach; it looks like it creates a MySqlDataSource instance for each and every instance of MySqlRelationalConnection; MySqlDataSource typically represents a connection pool and should be shared.

In general, an EF provider (e.g. the Pomelo provider) should not be concerned with configuring the lower-level ADO.NET provider (in this case the MySqlConnector). MySqlConnector itself should accept the ILoggingFactory - I suggest raising this on that repo.

bgrainger:
> The way to do this is to have the Pomelo provider accept an already-configured data source, and to have that used where a database connection is needed.

One potential issue with this is that it looks like Pomelo sometimes (or always? I haven't fully followed the code path) adds options to the connection string: https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/0209341b749f32109035f9b6955b7cddfd289218/src/EFCore.MySql/Storage/Internal/MySqlRelationalConnection.cs#L75

This wouldn't be possible when using a `MySqlDataSource` directly as the connection string for the data source is fixed. Perhaps the Pomelo code would need to detect unsupported options and throw an exception with instructions on how to fix the problem, rather than implicitly coercing them?

trejjam:
It could be done if `AddMySqlDataSource` is using Options pattern:

```cs
new ServiceDescriptor(
typeof(MySqlDataSource),
serviceProvider =>
{
	var dataSourceBuilder = new MySqlDataSourceBuilder(
                serviceProvider.GetService<IOptions<MySqlConnectionStringBuilder>>().Value.ConnectionString
        ).UseLoggerFactory(serviceProvider.GetService<ILoggerFactory>());
	dataSourceBuilderAction?.Invoke(dataSourceBuilder);
	return dataSourceBuilder.Build();
},
dataSourceLifetime));
```

EF then can do:
```cs
serviceCollection.AddOptions<MySqlConnectionStringBuilder>()
  .Configure<IOptions<PomeloEfCoreOptions>>((pomeloOptions, mysqlOptions) => mysqlOptions.TreatTinyAsBoolean = pomeloOptions.TreatTinyAsBoolean);

serviceCollection.AddSingleton<IValidateOptions<MySqlConnectionStringBuilder>, EfCoreMysqlValidation>();

public class EfCoreMysqlValidation : IValidateOptions<MySqlDataSourceOptions>
{
    public EfCoreMysqlValidation (IOptions<PomeloEfCoreOptions> pomeloEfCoreOptions) {
        _pomeloEfCoreOptions = pomeloEfCoreOptions;
    }

    public ValidateOptionsResult Validate(string? name, MySqlConnectionStringBuilder options)
    {
        // validate options.TreatTinyAsBoolean == _pomeloEfCoreOptions.Value.TreatTinyAsBoolean
    }
}
```

trejjam:
By that EF can set and also verify configuration (so a user does not shoot himself in the leg) and heavy lifting is done in the `MysqlConnector` library

lauxjpn:
> In general, an EF provider (e.g. the Pomelo provider) should not be concerned with configuring the lower-level ADO.NET provider (in this case the MySqlConnector). ~MySqlConnector itself should accept the ILoggingFactory - I suggest raising this on that repo.~ The way to do this is to have the Pomelo provider accept an already-configured data source, and to have that used where a database connection is needed.

@trejjam I agree with @roji.

---

> I've added DbDataSource specifically to EFCore.PG in the meantime, so you could copy that, but ideally this would all be managed in the core.

@roji Thanks, I'll take a look at it.

trejjam:
Closing in favor of #1817

PR #1827, opened 2024-01-08
Support custom JsonSerializerOptions in EFCore.MySql.Json.Microsoft

trejjam:
You can use source-generated JsonSerializerOptions or provide additional converters when using JSON columns. This PR makes it possible

### DIFF
--- a/src/EFCore.MySql.Json.Microsoft/Infrastructure/IMysqlJsonOptions.cs
+++ b/src/EFCore.MySql.Json.Microsoft/Infrastructure/IMysqlJsonOptions.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System.Text.Json;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Json.Microsoft.Infrastructure;
+
+public interface IMysqlJsonOptions
+{
+    JsonSerializerOptions JsonSerializerOptions { get; }
+}
--- a/src/EFCore.MySql.Json.Microsoft/Storage/ValueConversion/Internal/MySqlJsonMicrosoftPocoValueConverter.cs
+++ b/src/EFCore.MySql.Json.Microsoft/Storage/ValueConversion/Internal/MySqlJsonMicrosoftPocoValueConverter.cs
@@ -9,17 +9,17 @@ namespace Pomelo.EntityFrameworkCore.MySql.Json.Microsoft.Storage.ValueConversio
 {
     public class MySqlJsonMicrosoftPocoValueConverter<T> : ValueConverter<T, string>
     {
-        public MySqlJsonMicrosoftPocoValueConverter()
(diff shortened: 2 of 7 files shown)


src/EFCore.MySql.Json.Microsoft/Infrastructure/IMysqlJsonOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Pomelo Foundation. All rights reserved.
+// Licensed under the MIT. See LICENSE in the project root for license information.
+
+using System.Text.Json;
+
+namespace Pomelo.EntityFrameworkCore.MySql.Json.Microsoft.Infrastructure;
+
+public interface IMysqlJsonOptions
```
chullybun, line 8: This looks like a casing issue; shouldn't it be `IMySqlJsonOptions` for consistency?

trejjam: I will change it

src/EFCore.MySql.Json.Microsoft/Storage/ValueConversion/Internal/MySqlJsonMicrosoftPocoValueConverter.cs
```diff
@@ -9,17 +9,17 @@ namespace Pomelo.EntityFrameworkCore.MySql.Json.Microsoft.Storage.ValueConversio
...
 
-        private static string ConvertToProviderCore(T v)
-            => JsonSerializer.Serialize(v);
+        private static string ConvertToProviderCore(T v, JsonSerializerOptions jsonSerializerOptions)
+            => JsonSerializer.Serialize(v, jsonSerializerOptions);
 
-        private static T ConvertFromProviderCore(string v)
-            => JsonSerializer.Deserialize<T>(v);
+        private static T ConvertFromProviderCore(string v, JsonSerializerOptions jsonSerializerOptions)
+            => JsonSerializer.Deserialize<T>(v, jsonSerializerOptions);
```
lauxjpn, line 23: I am not sure at the moment, that we want to force the use of a `jsonSerializerOptions` parameter. We might want to make it optional, falling back on the default options.

src/EFCore.MySql.Json.Microsoft/Infrastructure/IMysqlJsonOptions.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    JsonSerializerOptions JsonSerializerOptions { get; }
+}
```
lauxjpn, line 11: I am not convinced that this is the right approach. I think users should have the ability to set `JsonSerializerOptions` using the model, not injecting them via a service.

They should be able to set a default on the model itself and then also concrete options on (JSON) properties of entities (that override the default options). They also need to be supplied in a JSON provider specific fashion (since we support Microsoft and Newtonsoft JSON implementations). So this requires JSON provider specific extension methods.

lauxjpn (changes requested):

PR #1832, opened 2024-01-18
Allow ExceptIntercept, ExceptInterceptPrecedence for MySql versions >= 8.0.31

jeffska:
As of version 8.0.31, MySql supports INTERSECT and EXCEPT. See https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-31.html

This change enables ExceptIntercept and ExceptInterceptPrecedence for versions >= 8.0.31.

The now-enabled tests pass, and it also works properly in my application.


jeffska:
Looks like failures are due to missing test overrides in unrelated ares.

lauxjpn:
@jeffska Thank you for your contribution!

PR #1850, opened 2024-02-27
Workaround SELECT ... ORDER BY (SELECT 1) related bug in Oracle's MySQL implementation, that can lead to completely wrong data being returned.

### LINKED ISSUE #1849: Using `SELECT ... ORDER BY (SELECT 1)` can lead to the return of unexpected data using Oracle MySQL releases
Using a `SELECT ... ORDER BY (SELECT 1)` expression can return completely wrong data using Oracle MySQL releases.
We came across this issue in our test suite, when tests are executed in a specific order.

The issue can be reproduced by executing a no-op (or any existing) test of the `ProxyGraphUpdatesMySqlTest.LazyLoading` class (e.g. our own `DummyTest`), immediately followed by `NorthwindSplitIncludeNoTrackingQueryMySqlTest.Include_collection_OrderBy_empty_list_contains(async: False)`.

We should implement a workaround, that outputs `1` instead of `(SELECT 1)`.

### Further technical details

MySQL version: 8.0.0+ (tested e.g. with 8.0.36)

lauxjpn:
The `QuerySqlGenerator` base implementation translates constant and parameter expressions in an `OrderingExpression` to `(SELECT 1)`.

This leads to a bug in Oracle's MySQL, where completely wrong data is being returned under certain conditions.

As a workaround, we now just output `1` instead of `(SELECT 1)` for all database versions and types.

Fixes #1849

### DIFF
--- a/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
+++ b/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
@@ -866,6 +866,28 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
             return mySqlBinaryExpression;
         }
 
+        protected override Expression VisitOrdering(OrderingExpression orderingExpression)
+        {
+            // The base implementation translates this case to `(SELECT 1)`.
+            // This leads to a bug in Oracle's MySQL, where completely wrong data is being returned under certain conditions.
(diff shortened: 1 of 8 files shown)


ChrisJollyAU:
@lauxjpn Interesting. Had to do the exact same for Jet

ChrisJollyAU:
@lauxjpn Actually Jet would not encounter the same problem. It doesn't support positional `ORDER BY`. Only normal column expressions. Had also forgotten that it had used just `SELECT 1` all the way from its early days. My thoughts were actually on a modification I had to do for when it was within an `WHERE NOT EXISTS` clause (https://github.com/CirrusRedOrg/EntityFrameworkCore.Jet/pull/202 )

Issue #1875, opened 2024-03-10
Change return type of all `EF.Functions.DateDiff` extension methods related to units of `SECOND` or smaller from `int` to `long`


lauxjpn:
The value returned for `TIMESTAMPDIFF` function calls of valid dates can exceed `Int32.MaxValue`, but currently, all our `EF.Functions.DateDiff` extension methods use `int` as the return type.

We should change this behavior to return `long` instead of `int`. At least for extension methods related to units of `SECOND` or smaller.

---

@ajcvickers The SQL Server provider currently uses `int` as the return type for everything as well, even though it translates those calls to `DATEDIFF_BIG`. An oversight or a conscious choice?

ajcvickers:
Probably an oversight. @roji?

roji:
As far as I can tell, the SQL Server EF.Functions.DateDiff* methods translate to DATEDIFF, not DATEDIFF_BIG ([code](https://github.com/dotnet/efcore/blob/main/src/EFCore.SqlServer/Query/Internal/SqlServerDateDiffFunctionsTranslator.cs#L561)); we seem to translate to DATEDIFF_BIG only when the user specifies DateTimeOffset.ToUnixTimeSeconds/ToUnixTimeMilliseconds. We could inrtoduce EF.Functions.DateDiffBig* overloads - see https://github.com/dotnet/efcore/issues/32278 which tracks this.

All this seems to be in line with the general policy that stuff in EF.Functions corresponds directly to SQL functions (as much as possible)... FWIW the Pomelo provider deviates from this in having EF.Functions.DateDiff rather than EF.Functions.TimestampDiff. In any case, SQL Server and MySQL seem to be different on this, with SQL Server having two functions (DATEDIFF returning int and DATEDIFF_BIG returning bigint), and MySQL having just one TIMESTAMPDIFF returning a bigint.

moander:
I vote for new `xBig` methods and leave todays `int` functions as is

lauxjpn:
@roji You're right, I did not look close enough.

As for Pomelo, unless we want to duplicate the extension methods and then mark all the old ones as obsolete (which we could arguably do), it would be a bit late in the game for us to rename them now to mimic the original MySQL function names.

We will probably leave it as is for now (its probably close enough), unless the community thinks a name change is worth it.
 
---

> I vote for new `xBig` methods and leave todays `int` functions as is

[...]

PR #1876, opened 2024-03-10
Add missing and convenient `EF.Functions.DateDiff` translations

### LINKED ISSUE #1837: Feature Request: DateDiff Function
Hello,

The MySQLDbFunctionsExtensions class has support for many of the specific date difference operations that MySql supports through its function TIMESTAMPDIFF(TIME_UNIT, startDate,endDate), but it doesn't support all of them.  For example, there is no support for using QUARTER as a time metric, which you can do in MySql.  Of course, the result for MONTH can be divided by 3, as a workaround, which is fine, but there is another issue:

There is no way to write one LINQ Query that takes a unit of measurement as a parameter to the TIMESTAMPDIFF function.  So you have to write multiple LINQ Queries in scenarios where you shouldn't have to.  One for dealing with weeks, a fixed interval, and another for months, which is a variable interval, for example.

Solution:

I request an added function, which takes as an additional parameter a string or an enumeration that represents a Time Unit, so that you could determine in advance of a LINQ query what the time interval unit will be.  As follows:

public static int DateDiff(**string unit,** this DbFunctions _, DateTime startDate, DateTime endDate)

Thanks,

lauxjpn:
Adds the following translations that are directly supported by MySQL compatible databases:
- `DateDiffQuarter`
- `DateDiffWeek`

Add the following convenience translations:
- `DateDiffMillisecond`
- `DateDiffTick`
- `DateDiffNanosecond`

#1875 will change the return type (at least of the second group) in the future from `int` to `long`.

Fixes #1837 (in part)

### DIFF
--- a/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
+++ b/src/EFCore.MySql/Extensions/MySqlDbFunctionsExtensions.cs
@@ -8,6 +8,8 @@ using Microsoft.EntityFrameworkCore.Diagnostics;
 // ReSharper disable once CheckNamespace
 namespace Microsoft.EntityFrameworkCore
 {
+    // TODO: Change method return types for units of `SECOND` and smaller from `int` to `long`.
+
     /// <summary>
     ///     Provides CLR methods that get translated to database functions when used in LINQ to Entities queries.
     ///     The methods on this class are accessed via <see cref="EF.Functions" />.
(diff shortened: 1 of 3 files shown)


lauxjpn:
> (From https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1875#issuecomment-1987951733:)
>
> All this seems to be in line with the general policy that stuff in EF.Functions corresponds directly to SQL functions (as much as possible) [...]

@roji Is there an existing/recommended pattern for adding convenience methods that the database does not implement (but we translate into working SQL anyway) and that complement methods in `EF.Functions`?

An example would be from the OP:

> Add the following convenience translations:
> 
> * `DateDiffMillisecond`
> * `DateDiffTick`
> * `DateDiffNanosecond`

roji:
Hmm, I'm not sure we have such convenience methods at the moment - I'd have to look.

OTOH I'd introduce these on EF.Functions alongside methods that translate directly to database functions - I'm not sure the user should care whether something translates directly to a SQL function or to a more elaborate construct.

BTW note that an alternative to providing an EF.Functions convenience thing, is to pattern-match a more complex construct in the query tree, i.e. where we avoid introducing a special function. That may be preferable - as long as it makes sense and allows expressing the same thing without a special thing...

PR #1899, opened 2024-03-21
Use JSON_VALUE() for scalar JSON values if available or workaround weird JSON_EXTRACT() behavior in regards to `null` otherwise

### LINKED ISSUE #1897: Replace JSON_EXTRACT utilisation with JSON_VALUE
### Steps to reproduce
Run the custom project to see it crashing
[PomeloBugJsonExtract.zip](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/files/14668199/PomeloBugJsonExtract.zip)

### The issue
In some places, in the code, `JSON_EXTRACT` is used for getting value of a JSON field when using projection or any other mean of directly getting data. But when the data is `null`, `JSON_EXTRACT` does return the string value `"null"` instead of the mysql `null` value, as intended in the spec : https://bugs.mysql.com/bug.php?id=85755
Sadly if you have a `Nullable<Datetime>` later in ef core when trying to convert it will use the string representation instead of the correct `null` value and crash with

```
---> System.FormatException: Couldn't interpret value as a valid DateTime: null
         at MySqlConnector.Core.Row.GetDateTime(Int32 ordinal) in /_/src/MySqlConnector/Core/Row.cs:line 336
         at MySqlConnector.MySqlDataReader.GetDateTime(Int32 ordinal) in /_/src/MySqlConnector/MySqlDataReader.cs:line 264
         at lambda_method186(Closure, QueryContext, DbDataReader, ResultContext, SingleQueryResultCoordinator)
         --- End of inner exception stack trace ---
         at lambda_method186(Closure, QueryContext, DbDataReader, ResultContext, SingleQueryResultCoordinator)
         at Microsoft.EntityFrameworkCore.Query.Internal.SingleQueryingEnumerable`1.Enumerator.MoveNext()
```

be aware that the `null` indication here is not a real null value but instead a string `"null"` in 

https://github.com/mysql-net/MySqlConnector/blob/1d508bcfd74f221e71a892bbb3086a545f1a5451/src/MySqlConnector/Core/Row.cs#L326-L342

### The naive solution ###
If we replace all call with `JSON_VALUE` we get a correct null value instead of a string and "everything" seems to work correctly

### Further technical details

MySQL version: 8.3.0-1.el8
Operating system:  Window11
Pomelo.EntityFrameworkCore.MySql version: 8.0.2

lauxjpn:
We will use `JSON_VALUE()` for scalar JSON values from now on, because it seems to handle `null` values correctly.

`JSON_EXTRACT()` will return a **string** with the contents of `null` for a JSON `null` value instead of a SQL `NULL` value (see [Bug #85755: JSON containing null value is extracted as a string "null"](https://bugs.mysql.com/bug.php?id=85755)).

So when we have to depend on `JSON_EXTRACT()`, we have to explicitly check for that specific case to ensure that the expected value is returned.

Fixes #1897

### DIFF
--- a/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
+++ b/src/EFCore.MySql/Query/ExpressionVisitors/Internal/MySqlQuerySqlGenerator.cs
@@ -106,13 +106,33 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionVisitors.Internal
 
         protected virtual Expression VisitJsonPathTraversal(MySqlJsonTraversalExpression expression)
         {
-            // If the path contains parameters, then the -> and ->> aliases are not supported by MySQL, because
-            // we need to concatenate the path and the parameters.
-            // We will use JSON_EXTRACT (and JSON_UNQUOTE if needed) only in this case, because the aliases
-            // are much more readable.
-            var isSimplePath = expression.Path.All(
-                l => l is SqlConstantExpression ||
-                     l is MySqlJsonArrayIndexExpression e && e.Expression is SqlConstantExpression);
+            var isScalar = expression.Type.IsPrimitive();
+            var useJsonValue = _options.ServerVersion.Supports.JsonValue &&
+                               isScalar;
+
+            if (useJsonValue)
+            {
+                VisitJsonPathTraversalCore(expression, "JSON_VALUE");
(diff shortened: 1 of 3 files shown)


crjc:
What steps are needed to move this PR forward? 
I’m also encountering the issue described in #1897 and interested in helping if I can.

PR #1906, opened 2024-04-14
Remove explicit reference to `System.Text.Json` package from `Pomelo.EntityFrameworkCore.MySql.Json.Microsoft` package

### LINKED ISSUE #1905: Remove STJ dependency as it is provided by the framework
**Is your feature request related to a problem? Please describe.**
I want to minimise dependencies in my project by utilising framework dependencies wherever possible

**Describe the solution you'd like**
I want the package to not have an explicit dependency on System.Text.Json as it can be provided by the framework. 

**Describe alternatives you've considered**
Accept the additional dependency

**Additional context**
n/a

thompson-tomo:
Removes the single reference to the  `System.Text.Json` package from the `Pomelo.EntityFrameworkCore.MySql.Json.Microsoft` package, because `System.Text.Json` has been an official part of .NET for quite a while and does not need to be explicitly references (anymore).

Closes #1905

### DIFF
--- a/Dependencies.targets
+++ b/Dependencies.targets
@@ -13,7 +13,6 @@
     <PackageReference Update="MySqlConnector.DependencyInjection" Version="2.3.5" />
 
     <PackageReference Update="NetTopologySuite" Version="2.5.0" />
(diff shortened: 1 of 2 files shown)


lauxjpn:
@thompson-tomo Thank you for your contribution!

Issue #1910, opened 2024-04-19
TimeSpan.TotalHours could not be translated


Lyra1337:
### Steps to reproduce
1. Create a Database:
```
CREATE DATABASE test1;
USE test1;
CREATE TABLE `test` (
	`Id` INT NOT NULL,
	`Time` TIME NULL,
	PRIMARY KEY (`Id`)
)
COLLATE='utf8mb4_general_ci';
```
2. Scaffold new context
3. Try to sum all values of the Time column

`context.Test.Sum(x => x.Time.TotalHours)`

### The issue

```
System.InvalidOperationException: The LINQ expression [...] could not be translated. Either rewrite the query in a form that can be translated, or switch to client evaluation explicitly by inserting a call to 'AsEnumerable', 'AsAsyncEnumerable', 'ToList', or 'ToListAsync'. See https://go.microsoft.com/fwlink/?linkid=2101038 for more information.
```
### Further technical details

Operating system: Windows / Linux
Pomelo.EntityFrameworkCore.MySql version: 8.0.0
Microsoft.AspNetCore.App version: 8.0.200

-----

So, it's dear to me that `TotalHours` can't be mapped.
Is there an other way to sum up all time valued-rows?
Can I add the function `TIME_TO_SEC` to the EF Functions to call it an then divide by 3600 to get the hours?

Any help will be appreciated.

Thank you in advance.

ajcvickers:
@lauxjpn Interesting you are planning on changing the `TimeSpan` mapping to `bigint`. We have the same problem with the mapping in SQL Server, and it may be that `bigint` is better, but it would be a break for existing Code First models. We're wondering if we should do the same thing, but concerned about the break. Any thoughts?

/cc @roji

lauxjpn:
@ajcvickers Pomelo has existing facilities in place to let users customize some type mappings (e.g. [MySqlDefaultDataTypeMappings](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/blob/main/src/EFCore.MySql/Infrastructure/MySqlDefaultDataTypeMappings.cs)), so users can do something like `.UseMySql([...], o => o.DefaultDataTypeMappings(m => m.WithClrTimeSpan(MySqlTimeSpanType.Time6)))`.

[...]

PR #1917, opened 2024-05-03
added translatons for TimeSpan.Total* members

Lyra1337:
Hello,
I tried to add support for TimeSpan.Total* members. I added two tests, I don't know how good the second test is - I can remove it or add another test with actual data.

Please tell me your thoughts about this code.

The idea about this sparked in #1910

### DIFF
--- a/src/EFCore.MySql/Query/Internal/MySqlMemberTranslatorProvider.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlMemberTranslatorProvider.cs
@@ -3,12 +3,15 @@
 
 using JetBrains.Annotations;
 using Microsoft.EntityFrameworkCore.Query;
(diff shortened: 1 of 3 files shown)


Lyra1337:
I just noticed that i accidentally reformatted the whole test file. I'll fix this in a minute.

Lyra1337:
I see there are some failed tests regarding to `DateTime.TotalDays` in `Pomelo.EntityFrameworkCore.MySql.FunctionalTests.Query.NorthwindFunctionsQueryMySqlTest.Datetime_subtraction_TotalDays`.
How do I specify that this translation is only targeting TimeSpan? Is checking `declaredType` in the `Translate` method not enough?

hlopez94:
You could investigate using https://github.com/nektos/act to locally try and run Github Actions on your PC or a Docker environment

PR #1937, opened 2024-08-06
Mark cloned connections as owned

trejjam:
Hi, I found out that when using DataSource, the produced connections are never disposed. Than can be solved by marking it as `owned: true`.


lauxjpn:
Looks good to me!

@trejjam Once you have rebased it on top of the current `master` branch, we can merge this one.

trejjam:
Done

lauxjpn:
@roji Npgsql might have this issue as well.

lauxjpn:
@trejjam Thank you for your contribution!

PR #1959, opened 2024-12-13
Support EF Core version targeted releases.

### LINKED ISSUE #1952: Traditional prerelease naming strategy can lead to discoverability issue when looking for the latest prerelease
> [...]
> Word of note for everyone: Visual Studio sorts prerelease releases by type. So it thinks that 9.0.0-preview.2.ci.20240426151710 is the latest prerelease version. You need to scroll down and look for 9.0.0-alpha.1.ci.20241207210245 in order to update to the latest version.

As pointed out by @MarkCiliaVincenti in https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/1944#issuecomment-2525549436_, there is a discoverability issue with our traditional naming strategy.

If we have already released a `preview` before (that targets the same EF Core preview) and then later produce prereleases independent of the EF Core prereleases (where we switch from `preview` to `alpha`, `beta` etc.), those newer prereleases get buried below the older `preview` prereleases when sorting by version in a descending fashion (see [SemVer2: Item 11](https://semver.org/#spec-item-11)).

Possible naming schemas could be:
- Instead of `9.0.0-alpha.1`, an alternative naming strategy could be `9.0.0-preview.alpha.1`.
- Just use the the `preview/rc/rtm/servicing` train independently of the EF Core one.
- Use the targeted EF Core version in the prelease string in addition to our own (or only add the `efcore...` part, if it is a uncorrelated release):
  - `9.0.0-preview.3.efcore.9.0.0`
  - `9.0.0-preview.3.efcore.9.0.0.preview.2`

The `alpha/beta/silver` train seems problematic in general, because silver is already ordered _after_ `rtm` and `servicing`.

lauxjpn:
Allows us to put the targeted EF Core version into the release string.

Fixes #1952

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -350,11 +350,17 @@ jobs:
               $configurations = $combination[1]
               foreach ($configuration in $configurations)
               {
-                $arguments = 'pack', '-c', $configuration, '-o', "nupkgs/$configuration/$type", '-p:ContinuousIntegrationBuild=true'
+                $arguments = 'pack', '-c', $configuration, '-o', "nupkgs/$configuration/$type", '-p:ContinuousIntegrationBuild=true', '-p:OutputVersionProperties=true'
 
                 if ($officialBuild)
                 {
-                  $arguments += "-p:OfficialVersion=$officialVersion"
+                  $finalOfficialVersion = $configuration -eq 'Debug' `
+                    ? $officialVersion.Contains('-') `
+                      ? $officialVersion + '.debug' `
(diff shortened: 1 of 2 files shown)


Supermenwild:
pomelo.entityframeworkcore.mysql.8.0.2

lauxjpn:
> pomelo.entityframeworkcore.mysql.8.0.2

@Supermenwild I have no idea what you mean by that. Care to elaborate?

PR #1988, opened 2025-03-06
Fix issue where using `DateTime.UtcNow` and `DateTime.Now` will lose precision when used inside of a query

### LINKED ISSUE #1987: DateTime.UtcNow Translates to UTC_TIMESTAMP(), Causing Precision Loss in Date Comparisons
**Describe the bug**  
When using `DateTime.UtcNow` inside an Entity Framework query with Pomelo.EntityFrameworkCore.MySql, it is translated into `UTC_TIMESTAMP()` in MySQL. This causes unexpected behavior because MySQL implicitly converts the compared datetime values to `seconds` precision, even if the original datetime had higher precision (e.g., milliseconds or microseconds).  

As a result, comparisons may yield incorrect results, such as:  
- Two datetimes appearing equal when they are actually different by milliseconds.  
- A datetime appearing older or newer than `DateTime.UtcNow` when it is actually the same.  

**To Reproduce**  

1. Create an entity with a `DateTime` property, e.g.:  
   ```csharp
   public class MyEntity  
   {  
       public int Id { get; set; }  
       public DateTime CreatedAt { get; set; }  
   }
   ```  
2. Insert an entity with a precise timestamp:  
   ```csharp
   var entity = new MyEntity { CreatedAt = DateTime.UtcNow };  
   dbContext.MyEntities.Add(entity);  
   dbContext.SaveChanges();  
   ```  
3. Run a query comparing `CreatedAt` to `DateTime.UtcNow`:  
   ```csharp
   var results = dbContext.MyEntities  
       .Where(e => e.CreatedAt <= DateTime.UtcNow)  
       .ToList();
   ```  
4. Observe that the results may be incorrect due to precision loss when MySQL compares `UTC_TIMESTAMP()` (seconds precision) with `CreatedAt` (higher precision).  

**Expected behavior**  
- `DateTime.UtcNow` should be translated into a MySQL function that preserves full precision, such as `UTC_TIMESTAMP(6)` or another suitable function.  
- Comparisons should work correctly, reflecting the actual precision of stored datetime values.  

**Technical details (please complete the following information):**

- Database server version: MySQL 8.0.41, running inside Docker container with image `mysql:8.4`.
- Operating system: Apple M1 Pro, Sequoia 15.2
- Pomelo.EntityFrameworkCore.MySql version: 8.0.3

**Additional context**  
- This issue can lead to subtle and hard-to-debug problems in applications that rely on precise datetime comparisons.  
- A workaround might involve storing `DateTime.UtcNow` to a variable outside of the query and then using it as that stores the precision correctly.

jeknom:
Fixes #1987

Hope this helps out!

### DIFF
--- a/src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
@@ -101,13 +101,13 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.Internal
                             declaringType == typeof(DateTimeOffset)
                                 ? "UTC_TIMESTAMP"
                                 : "CURRENT_TIMESTAMP",
-                            Array.Empty<SqlExpression>(),
+                            new [] { _sqlExpressionFactory.Constant(6)},
                             returnType);
 
                     case nameof(DateTime.UtcNow):
                         return _sqlExpressionFactory.NonNullableFunction(
                             "UTC_TIMESTAMP",
-                            Array.Empty<SqlExpression>(),
+                            new [] { _sqlExpressionFactory.Constant(6)},
                             returnType);
 
                     case nameof(DateTime.Today):
--- a/src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
+++ b/src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
@@ -8,6 +8,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Diagnostics;
 using Microsoft.EntityFrameworkCore.Query;
 using Microsoft.EntityFrameworkCore.Query.SqlExpressions;
+using Pomelo.EntityFrameworkCore.MySql.Infrastructure.Internal;
 using Pomelo.EntityFrameworkCore.MySql.Utilities;
 
 namespace Pomelo.EntityFrameworkCore.MySql.Query.Internal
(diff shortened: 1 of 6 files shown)


src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
```diff
@@ -101,13 +101,13 @@ public virtual SqlExpression Translate(
...
                                 : "CURRENT_TIMESTAMP",
-                            Array.Empty<SqlExpression>(),
+                            new [] { _sqlExpressionFactory.Constant(6)},
                             returnType);
 
                     case nameof(DateTime.UtcNow):
                         return _sqlExpressionFactory.NonNullableFunction(
                             "UTC_TIMESTAMP",
-                            Array.Empty<SqlExpression>(),
+                            new [] { _sqlExpressionFactory.Constant(6)},
```
lauxjpn, line 110: The general implementation looks good to me. However, we should probably ensure that a datetime precision is supported. While all major current database implementations support that by now, we try to stay compatible even with not officially supported versions as good as we can. Since we have checked for datetime precision support in the past in other places, we should keep doing it for now here as well.

With an `IMySqlOptions _options` field, you can check for `_options.ServerVersion.Supports.Datetime6`. The constructor of the `MySqlDateTimeMemberTranslator` currently does not take an `IMySqlOptions` parameter yet, so you have to add it and save it to a new field.

`MySqlDateTimeMemberTranslator` is instatiated by `MySqlMemberTranslatorProvider` and its constructor is the one that can get dependency injected services. So add a `IMySqlOptions` parameter to the `MySqlMemberTranslatorProvider` constructor fist, then add it to the constructor call of `MySqlDateTimeMemberTranslator`, save it to a field and finally use it in your code.

lauxjpn (changes requested):
Please search & replace `UTC_TIMESTAMP()` with `UTC_TIMESTAMP(6)` in the SQL baselines of the tests, so that they all work again.

lauxjpn:
The precision that MySQL supports (microseconds) and the one that .NET supports (100 nanoseconds) still doesn't fully match, but its definitely closer after this PR has been merged than before.

src/EFCore.MySql/Query/Internal/MySqlDateTimeMemberTranslator.cs
jeknom: Nice! 🙌 I added the check for Datetime6 support.

jeknom:
I fixed the tests now

lauxjpn (approved):
LGTM!

lauxjpn:
@jeknom Thank you for your contribution!

PR #2009, opened 2025-08-28
Do not force json to use JSON functions

Tobbeman:
We are using scaffolding to create our database classes. Sadly, the scaffold does not support json on the json field, instead opting for string.

This change will let the user use JSON functions more freely

### DIFF
--- a/src/EFCore.MySql/Query/ExpressionTranslators/Internal/MySqlJsonDbFunctionsTranslator.cs
+++ b/src/EFCore.MySql/Query/ExpressionTranslators/Internal/MySqlJsonDbFunctionsTranslator.cs
@@ -145,13 +145,7 @@ namespace Pomelo.EntityFrameworkCore.MySql.Query.ExpressionTranslators.Internal
 
             return result;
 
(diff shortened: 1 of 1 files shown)


lara-ec:
In case this is helpful for anyone: our solution for this was editing the `EntityType.t4` template to not annotate the `string` columns as `json`:

before:
```t4
            foreach (var dataAnnotation in dataAnnotations)
            {
#>
    <#= code.Fragment(dataAnnotation) #>
<#
            }
```

after:
```t4
            foreach (var dataAnnotation in dataAnnotations)
            {
                var annotation = dataAnnotation;
                if (
                    dataAnnotation.Type == typeof(System.ComponentModel.DataAnnotations.Schema.ColumnAttribute)
                    && dataAnnotation.NamedArguments.TryGetValue("TypeName", out var typeName)
                    && typeName is "json"
                )
                {
                    annotation = new AttributeCodeFragment(
                        annotation.Type,
                        annotation.Arguments,
                        annotation.NamedArguments.Where(n => n.Key != "TypeName").ToDictionary()
                    );

#>
    // this is a json column
<#
                }
#>
    <#= code.Fragment(annotation) #>
<#
            }
```

PR #2026, opened 2026-01-07
.Net 10 changes for reference

renemadsen:
@lauxjpn Use for refference as you wish.

Let me know in what slices you want the changes to be put into different smaller PR's for you to better use the changes.
I've kept all the existing db versions and added new versions to demonstrate that no regressions are there.

To my best of knowledge and testing, all changes are working.
Also implemented ToJson() instead of the custom json implementation, since I personally don't like custom out of scope changes => more custom code to maintain, than following the mainline => easier to maintain.


ITaluone:
Hello @renemadsen 

Just want to confirm: with our most complex app all tests are green (local and CI); also manual testing works...

Just updated to `Microsoft.EntityFrameworkCore.*` to 10.0.1 and your preview.3 package everywhere.. Nothing more needed

Apps specifics:
- ASP.NET Core 10
- Mariadb 11.8

Performance is the same as with EF Core 9

Thank you for you work!

PR #2027, opened 2026-01-10
Fix migrations outputting static method calls instead of fluent chaining

### LINKED ISSUE #1990: `..BuilderExtension` methods are called as `...BuilderExtensions.Method()` instead of `.Method()` in migrations
Currently, extension methods are called in migrations as in the following example:

```c#
modelBuilder.Entity("IssueConsoleTemplate.IceCream", b =>
    {
        b.Property<int>("IceCreamId")
            .ValueGeneratedOnAdd()
            .HasColumnType("int");

        MySqlPropertyBuilderExtensions.UseMySqlIdentityColumn(b.Property<int>("IceCreamId"));

        b.Property<string>("Name")
            .HasColumnType("longtext");

        MySqlPropertyBuilderExtensions.HasCharSet(b.Property<string>("Name"), "latin1");
    }):
```

A better representation would be the following one (which I believe has worked in some earlier release trains):

```c#
modelBuilder.Entity("IssueConsoleTemplate.IceCream", b =>
    {
        b.Property<int>("IceCreamId")
            .ValueGeneratedOnAdd()
            .HasColumnType("int")
            .UseMySqlIdentityColumn("IceCreamId");

        b.Property<string>("Name")
            .HasColumnType("longtext")
            .HasCharSet("latin1");
    }):
```

- Pomelo.EntityFrameworkCore.MySql version: at least in 9.0 (need to check other release trains)

kimjaejung96:
Use method name strings instead of MethodInfo for MethodCallCodeFragment construction. This ensures EF Core's CSharpSnapshotGenerator outputs fluent chaining (e.g., .HasCharSet("latin1")) instead of static method calls (e.g., MySqlPropertyBuilderExtensions.HasCharSet(...)).

Add unit tests to verify MethodInfo is null for fluent API fragments.

Fixes #1990


kimjaejung96:
Hi! @lauxjpn I’ve opened PR #2027 to address #1990. :)
The change makes EF Core generate fluent chaining by using method name strings for MethodCallCodeFragment, and includes tests verifying MethodInfo is null for the relevant fragments.

PR #2049, opened 2026-09-07
Add .NET 10 support, row-level locking and UPSERT extensions

mmqiu123456:
* Add .NET 10 target framework support
* Add row-level locking extensions (ForUpdate, ForShare, ForUpdateNoWait, ForUpdateSkipLocked) emitting FOR UPDATE / FOR SHARE clauses
* Add UPSERT extensions (UpsertAsync, UpsertRangeAsync) backed by INSERT ... ON DUPLICATE KEY UPDATE
* Adapt query compilation context and parameter-based SQL processor for EF Core 10 API surface
* Keep Version.props at 9.0.0 (versioning left to maintainers)


mmqiu123456:
## CI status

All 19 checks are now green (18 success + 1 skipped).

The 12 MariaDB jobs were failing in the first CI run; this is fixed by commit `3a5eb9e`.

## Why MariaDB is excluded

The two new features emit MySQL 8.0+ only syntax:

1. **Row-level locking** — `FOR SHARE` is MySQL 8.0+ syntax. MariaDB only provides the legacy `LOCK IN SHARE MODE` and does not accept `FOR SHARE`.
   Note that `FOR UPDATE`, `NOWAIT` and `SKIP LOCKED` *are* supported by MariaDB 10.6+ (those tests passed on MariaDB already).

2. **UPSERT** — the generated statement uses a row alias, `INSERT ... AS new ON DUPLICATE KEY UPDATE ...`, which requires MySQL 8.0.19+. MariaDB does not support row aliases at all.

Since MariaDB cannot parse either construct, `MySqlLockingAndUpsertExtensionsTest` is now annotated with `[SupportedServerVersionCondition("8.0.19")]`, which skips the class on MariaDB and on MySQL < 8.0.19.

## Scope

- **MySQL 8.0.19+**: both features are fully covered by tests.
- **MariaDB**: features unavailable — this is a server-side syntax limitation, not a provider limitation.

Happy to adjust the approach if you would prefer the provider to emit MariaDB-compatible SQL instead (e.g. `LOCK IN SHARE MODE` on MariaDB, and a non-aliased `ON DUPLICATE KEY UPDATE`). That is a larger change touching SQL generation, so I kept this PR minimal and aligned with the existing `SupportedServerVersionCondition` convention.

